PR #41, opened 2017-05-16
Using topbraid

mandawilson:
We still need to:

1) Need better error handling, especially for when we fail to authenticate to TopBraid.  This is what we get right now:

Could not extract response: no suitable HttpMessageConverter found for response type [java.util.List<org.mskcc.oncotree.topbraid.OncoTreeNode>] and content type [text/html;charset=utf-8]

2) Add versions (the API had support for it, I ignore it)
3) If parent not found for a child it logs an error, needs to deal with the error in some way
4) I had to comment out "level" query filtering, we probably need to add that back at some point
5) "level" and "tissue" fields are always null, we probably also need that back

I added log4j so that I could quickly get logging, but we may not want to do this.  

I threw all the Topbraid stuff from pipelines into one package.  I am happy to do something else, again just wanted something working quickly.

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeNode.java
+++ b/core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeNode.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright (c) 2017 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
+ * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
+ * is on an "as is" basis, and Memorial Sloan-Kettering Cancer Center has no
+ * obligations to provide maintenance, support, updates, enhancements or
+ * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
+ * liable to any party for direct, indirect, special, incidental or
+ * consequential damages, including lost profits, arising out of the use of this
+ * software and its documentation, even if Memorial Sloan-Kettering Cancer
+ * Center has been advised of the possibility of such damage.
+ */
+
+package org.mskcc.oncotree.topbraid;
+
+/**
+ *
+ * @author Manda Wilson
+ */
+
+import java.util.HashMap;
+import java.util.Map;
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+    "code",
+    "name",
+    "mainType",
+    "color",
+    "nci",
+    "umls",
+    "nccn",
+    "parentCode"
+})
+public class OncoTreeNode {
+
+    @JsonProperty("code")
+    private String code;
+
+    @JsonProperty("name")
+    private String name;
+
+    @JsonProperty("mainType")
+    private String mainType;
+
+    @JsonProperty("color")
+    private String color;
+
+    @JsonProperty("nci")
+    private String nci;
+
+    @JsonProperty("umls")
+    private String umls;
+
+    @JsonProperty("nccn")
+    private String nccn;
+
+    @JsonProperty("parentCode")
+    private String parentCode;
+
+    @JsonIgnore
+    private Map<String, Object> additionalProperties = new HashMap<String, Object>();
+
+    @JsonProperty("code")
+    public String getCode() {
+        return code;
+    }
+
+    @JsonProperty("code")
+    public void setCode(String code) {
+        this.code = code;
+    }
+
+    @JsonProperty("name")
+    public String getName() {
+        return name;
+    }
+
+    @JsonProperty("name")
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    @JsonProperty("mainType")
+    public String getMainType() {
+        return mainType;
+    }
+
+    @JsonProperty("mainType")
+    public void setMainType(String mainType) {
+        this.mainType = mainType;
+    }
+
+    @JsonProperty("color")
+    public String getColor() {
+        return color;
+    }
+
+    @JsonProperty("color")
+    public void setColor(String color) {
+        this.color = color;
+    }
+
+    @JsonProperty("nci")
+    public String getNci() {
+        return nci;
+    }
+
+    @JsonProperty("nci")
+    public void setNci(String nci) {
+        this.nci = nci;
+    }
+
+    @JsonProperty("umls")
+    public String getUmls() {
+        return umls;
+    }
+
+    @JsonProperty("umls")
+    public void setUmls(String umls) {
+        this.umls = umls;
+    }
+
+    @JsonProperty("nccn")
+    public String getNccn() {
+        return nccn;
+    }
+
+    @JsonProperty("nccn")
+    public void setNccn(String nccn) {
+        this.nccn = nccn;
+    }
+
+    @JsonProperty("parentCode")
+    public String getParentCode() {
+    return parentCode;
+    }
+
+    @JsonProperty("parentCode")
+    public void setParentCode(String parentCode) {
+        this.parentCode = parentCode;
+    }
+
+    @JsonAnyGetter
+    public Map<String, Object> getAdditionalProperties() {
+        return this.additionalProperties;
+    }
+
+    @JsonAnySetter
+    public void setAdditionalProperty(String name, Object value) {
+        this.additionalProperties.put(name, value);
+    }
+
+}
--- a/core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeRepository.java
+++ b/core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeRepository.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2017 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.  The software and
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeNode.java
```diff
@@ -0,0 +1,162 @@
...
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({
+    "code",
+    "name",
+    "mainType",
+    "color",
+    "nci",
+    "umls",
+    "nccn",
```
sheridancbio: There will be no nccn with the oncotree we are currently releasing .. also we are adding ICDO. Maybe we should keep oncotree nodes flexible and represent all of these kinds of attributes an non-named "additionalAttributes" in the model somehow. For now, we can leave this model as is.

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeRepository.java
```diff
@@ -0,0 +1,64 @@
...
+    private String username;
+
+    @Value("${topbraid.oncotree.password:}")
+    private String password;
+    
+    private String query = "PREFIX skos:<http://www.w3.org/2004/02/skos/core#> " +
+        "PREFIX onc:<http://data.mskcc.org/ontologies/oncotree/> " +
+        "SELECT DISTINCT ?code ?name ?mainType ?color ?nci ?umls ?icdo ?parentCode " +
+        "WHERE { " +
+        " GRAPH <urn:x-evn-master:oncotree_2017_05_16> { " +
```
sheridancbio: We want to parameterize this .. but for now, fine.

core/src/main/java/org/mskcc/oncotree/topbraid/TopBraidRepository.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private final static Logger logger = Logger.getLogger(TopBraidRepository.class);
+
+    @Value("${topbraid.url}")
+    private String topBraidURL;
+
+    @Autowired
+    private TopBraidSessionConfiguration topBraidSessionConfiguration; 
+
+    protected List<T> query(String username, String password, String query, ParameterizedTypeReference parameterizedType) {
```
sheridancbio: Here we are using the topbraid.url string through spring value setting, but passing in the username and password as caller parameters. I'm not sure why we use different mechanisms to pass in config info.

core/src/main/java/org/mskcc/oncotree/topbraid/TopBraidSessionConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+ **/
+@Configuration
+public class TopBraidSessionConfiguration {
+
+    private final static Logger logger = Logger.getLogger(TopBraidSessionConfiguration.class);
+
+    @Value("${topbraid.url}")
+    private String topBraidURL;
+    
+    public String getSessionId(String username, String password) {
```
sheridancbio: Same comment as before .. not sure why url comes in through spring value setting and username/pw come in through function parameter passing.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -20,32 +26,22 @@
...
 
-            tumorTypes = parseFromRaw(new FileInputStream(properties.getProperty("tumor_type_file_path")), version);
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
-        }
-        return tumorTypes;
-    }
+    private static OncoTreeRepository oncoTreeRepository;
+    @Autowired
+    public void setOncoTreeRepository(OncoTreeRepository property) { oncoTreeRepository = property; }
```
sheridancbio: I'm not familiar with autowiring a method definition. I'm not sure I know what this means.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -20,32 +26,22 @@
...
+
+    private static final String PROPERTY_FILE = "classpath:application.properties";
+    private static List<String> TumorTypeKeys = Arrays.asList("code", "name", "nci", "level", "umls", "maintype", "color");
```
sheridancbio: So here we are leaving the model used by the frontend intact and filling the (frontend) model from what we get from our repository. This is ok. I would recommend that if we have different models in the repository and in the web layer, that we do model translation in a service layer (rather than in the repository). We can move towards that in a future PR

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -137,33 +136,60 @@ public static InputStream getTumorTypeInputStreamFromGitHub(Version version) {
...
-        //the line separator sequence is defined here to ensure systems such as MacOS and Windows
-        //are able to process this file correctly (MacOS uses '\r'; and Windows uses '\r\n').
-        settings.getFormat().setLineSeparator("\n");
-
-        // creates a TSV parser
-        TsvParser parser = new TsvParser(settings);
-
-        Map<String, TumorType> tumorTypes = new HashMap<>();
+        TumorType root = new TumorType();
+        root.setCode("TISSUE");
```
sheridancbio: We manually create the tissue node?  I think that it will be returned from topbraid. Maybe it currently is not showing up because the sparql query is requiring a parent (broader) link. But I think that in topbraid there *is* actually a parent for the tissue node .. called "OncotreeConcept" or something like that.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -137,33 +136,60 @@ public static InputStream getTumorTypeInputStreamFromGitHub(Version version) {
...
+        }
 
-        //Iterate each row and assign tumor type to parent following the order of appearing
-        for (String[] row : allRows.subList(1, allRows.size())) {
-            tumorType.setChildren(attachTumorType(tumorType.getChildren(), row, 0, version, "TISSUE"));
+        // now we have all nodes, fill in children
+        for (TumorType tumorType : allNodes.values()) {
+            // TISSUE is root and has no parent, skip
+            if (!tumorType.getCode().equals("TISSUE")) {
+                if (tumorType.getParent() == null) {
```
sheridancbio: From topbraid I would expect that the parent of primary nodes would come in as "TISSUE" .. if not, I think that is an error in the topbraid data loading.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -137,33 +136,60 @@ public static InputStream getTumorTypeInputStreamFromGitHub(Version version) {
...
+                    logger.debug("Parent is null for tumor type code '" +
+                        tumorType.getCode() + "'.  Adding to 'TISSUE' node.");
+                    tumorType.setParent(root.getCode());
+                    root.addChild(tumorType);
```
sheridancbio: right .. constructing all parent and child relationships based on the parent. This makes sense.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -137,33 +136,60 @@ public static InputStream getTumorTypeInputStreamFromGitHub(Version version) {
...
-        tumorTypes.put("TISSUE", tumorType);
+        return tree;
+    }
 
-        return tumorTypes;
+    private static TumorType initTumorType(OncoTreeNode oncoTreeNode, Version version) {
+        // we do not have level or tissue
+        TumorType tumorType = new TumorType();
+        if (oncoTreeNode.getMainType() != null) {
+            tumorType.setMainType(MainTypesUtil.getOrCreateMainType(oncoTreeNode.getMainType(), version));
```
sheridancbio: I wonder how maintype would be "created" if it is not known. For us, I think we may not want to guess and want to throw an exception if the mainType is not defined .. or leave it as undefined. Some of the primary level nodes have a (as of yet) undefined maintype and we plan to show those with cancer_type "NA" in the portal interface

core/src/main/java/org/mskcc/oncotree/topbraid/TopBraidSessionConfiguration.java
mandawilson: This is because the URL is general to TopBraid and the username and password is specific to Oncotree.  The OncotreeRepository has the username and parameter set by Spring and passes it to the parent TopBraidRepository.  You had said we might have different usernames and passwords for different things within TopBraid.  This is sort of unnecessary since this is the Oncotree app, rather than the planned generic TopBraid app, but since the code is already designed to plugin another [NewType]Repository, I would leave it ready to do that (ready to be pulled out into a separate web API).

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
mandawilson: Spring will magically call setOncoTreeRepository with an initialized OncoTreeRepository.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
mandawilson: I was not getting a TISSUE node in the query results.  The previous code also created this root TISSUE node.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
mandawilson: This creates the main type only if we get one from TopBraid.  If TopBraid does not have it the result returned is a TumorType with no mainType set.  In fact, in the previous code nothing from the primary column has a main type.

core/src/main/java/org/mskcc/oncotree/topbraid/TopBraidSessionConfiguration.java
sheridancbio: Thanks for explanation

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
sheridancbio: I guess this will happen on instantiation of the class. I thought that an Autowired class member (putting @Autowired before the member declaration for oncoTreeRepository) would accomplish something similar. But I guess the set method might have additional logic that you want to happen.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
sheridancbio: The spreadsheet had no tissue node, but if TISSUE is a valid oncotree code (maybe we need to confirm this), then it could/should have meta information (color, mainType) which can be set inside of topbraid and which we should try to extract and use.

PR #44, opened 2017-05-23
Gets and uses TopBraid versions

mandawilson:
Gets and uses TopBraid versions

Error handling

Set level and tissue

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/utils/VersionUtil.java
+++ b/core/src/main/java/org/mskcc/oncotree/utils/VersionUtil.java
@@ -1,41 +1,44 @@
 package org.mskcc.oncotree.utils;
 
 import org.mskcc.oncotree.model.Version;
+import org.mskcc.oncotree.error.InvalidVersionException;
+import org.mskcc.oncotree.topbraid.OncoTreeVersionRepository;
 
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
 
 /**
  * Created by Hongxin on 2/25/16.
  */
+@Component
 public class VersionUtil {
-    private static final Map<String, Version> VERSIONS = new HashMap<String, Version>() {{
-        put("1", new Version("1", "369b74c599ebefdb71bb25a85cb877be954a0928"));
-        put("1.1", new Version("1.1", "52a743f6a3493d1cb46eca2a3e12e7f92225558d"));
-        put("oncokb", new Version("oncokb", "a41b4b38aeabbcf0ae2a6414ee90add1fd72468d"));
-        put("genie", new Version("genie", "23fac821a946a3973b65d91b0616136222578e40"));
-        put("realtime", new Version("realtime", "realtime"));
-    }};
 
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/mskcc/oncotree/utils/VersionUtil.java
```diff
@@ -1,41 +1,44 @@
...
-
-    public static Version getVersion(String version) {
-        return VERSIONS.get(version);
-    }
 
-    public static Version getVersionOrRealtime(String version) {
-        return VERSIONS.get(version) == null ? VERSIONS.get("realtime") : VERSIONS.get(version);
+    private static OncoTreeVersionRepository oncoTreeVersionRepository;
+    @Autowired
+    public void setOncoTreeVersionRepository(OncoTreeVersionRepository property) { oncoTreeVersionRepository = property; }
```
sheridancbio: property seems like an odd name for the argument. "repository" or "value" would have been my thought, but maybe the value comes from a spring property?

core/src/main/java/org/mskcc/oncotree/utils/VersionUtil.java
```diff
@@ -1,41 +1,44 @@
...
+        return oncoTreeVersionRepository.getOncoTreeVersions();
     }
 
-    public static Set<Version> getVersions() {
-        Set<Version> versions = new HashSet<>();
-        for (Map.Entry<String, Version> entry : VERSIONS.entrySet()) {
-            versions.add(entry.getValue());
+    public static Version getVersion(String version) throws InvalidVersionException {
+        if (version != null && version.trim() != "") {
+            for (Version v : getVersions()) {
```
sheridancbio: This "see if a version exists" functionality will probably be needed elsewhere. Maybe it should be its own function (in oncoTreeVersionRepository). Also, with a long list of versions it may be a waste to iterate through them anyway. Perhaps if the repository knows the version that is being sought, it can do a sparql query which only asks for that single version during the query to TopBraid

core/src/main/java/org/mskcc/oncotree/utils/VersionUtil.java
```diff
@@ -1,41 +1,44 @@
...
-        return versions;
+        throw new InvalidVersionException("'" + version + "' is not a valid version.");
     }
 
-    public static Boolean hasVersion(String version) {
-        return VERSIONS.containsKey(version);
-    }
+    public static Version getDefaultVersion() throws InvalidVersionException {
+        // note we will throw an InvalidVersionException if this is not found in TopBraid
+        return getVersion("oncotree_current");
```
sheridancbio: this should probably be a configuration parameter (topbraid.oncotree.default_version)

web/src/main/java/org/mskcc/oncotree/api/MainTypesApi.java
```diff
@@ -46,7 +46,7 @@
         meta.setCode(200);
         resp.setMeta(meta);
 
-        Version v = VersionUtil.getVersionOrRealtime(version);
+        Version v = (version == null) ? VersionUtil.getDefaultVersion() : VersionUtil.getVersion(version);
```
sheridancbio: Ok using the trinary operator (a ? b : c) .. but if the expression was any more complex I might prefer an if/then

PR #46, opened 2017-05-30
Added unit testing and made other adjustments

sheridancbio:
- unit test class created (Mocks OncoTreeRepository, tests TumorTypeUtil)
    - all current oncotree nodes tested in output to getOncoTree(version)
- adjusted sparql query to return TISSUE node (without a parent ref)
- validated repostiory response before constructing TissueTypes
    - throw Exception if retrieved oncotree is invalid (from TopBraid)
    - tree must have exactly on root
    - all nodes must have a defined parent (except root)
    - root id not hardcoded : it is the node without parent
    - no oncoTreeCode is duplicated in tree
- added test config class for spring based test runner (not used yet)

### DIFF
--- a/core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
+++ b/core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
@@ -1,0 +1,887 @@
+/** Copyright (c) 2017 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.  The software and
+ * documentation provided hereunder is on an "as is" basis, and
+ * Memorial Sloan-Kettering Cancer Center
+ * has no obligations to provide maintenance, support,
+ * updates, enhancements or modifications.  In no event shall
+ * Memorial Sloan-Kettering Cancer Center
+ * be liable to any party for direct, indirect, special,
+ * incidental or consequential damages, including lost profits, arising
+ * out of the use of this software and its documentation, even if
+ * Memorial Sloan-Kettering Cancer Center
+ * has been advised of the possibility of such damage.
+*/
+
+package org.mskcc.oncotree.utils;
+
+import java.io.*;
+import java.nio.file.*;
+import java.util.*;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.runner.RunWith;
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -0,0 +1,866 @@
...
+            TumorType returnedTumorType = returnedTumorTypeMap.get(oncoTreeCode);
+            if (expectedTumorType == null || returnedTumorType == null) {
+                if (expectedTumorType == null) {
+                    failureReport.append(oncoTreeCode + " : no values in expected tumor type hash (internal error)\n");
+                }
+                if (returnedTumorType == null) {
+                    failureReport.append(oncoTreeCode + " : tumor type expected but not returned from tumorTypesUtil\n");
+                }
+            } else {
+                if (!expectedTumorType.getCode().equals(returnedTumorType.getCode())) {
```
mandawilson: These throw null pointer exceptions if code, name, color, etc is not set.

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -0,0 +1,866 @@
...
+                    failureReport.append(makeMismatchMessage(oncoTreeCode, "code", returnedTumorType.getCode(), expectedTumorType.getCode()));
+                    failureCount = failureCount + 1;
+                }
+                if (!expectedTumorType.getName().equals(returnedTumorType.getName())) {
+                    failureReport.append(makeMismatchMessage(oncoTreeCode, "name", returnedTumorType.getName(), expectedTumorType.getName()));
+                    failureCount = failureCount + 1;
+                }
+                if (!expectedTumorType.getColor().equals(returnedTumorType.getColor())) {
```
mandawilson: I changed getColor() in OncoTreeNode.java to return "blue" and successfully failed the tests:

Failed tests: 
  TumorTypesUtilTest.testFullOncoTreeCodeListByCode:212 644 failed test conditions. Details follow... HGSOS : mismatch in color (got:blue) (expected:White)
TET : mismatch in color (got:blue) (expected:Purple)
BLSC : mismatch in color (got:blue) (expected:Yellow)
RAS : mismatch in color (got:blue) (expected:LightYellow)
SCT : mismatch in color (got:blue) (expected:LightBlue)
HGNET : mismatch in color (got:blue) (expected:Gray)

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
sheridancbio: I have fixed this, so that null values in expected nodes are handled properly. (whitespace only string is equivalent to null)

PR #51, opened 2017-06-26
Connected tumor_types.txt api to topbraid

zheins:
Tumor_types.txt endpoint now gets data from topbraid instead of file on filesystem. Response is parsed into a tsv. Front end currently uses the tumot_types.txt api endpoint to render tree visualization, so this connects the front end to topbraid.

Not currently supporting NCI/UMLS fields.

@sheridancbio @mandawilson

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
+++ b/core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
@@ -23,6 +23,7 @@ import java.net.URL;
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import org.springframework.core.io.InputStreamResource;
 
 /**
  * Created by Hongxin on 2/25/16.
@@ -31,6 +32,7 @@ import java.util.regex.Pattern;
 public class TumorTypesUtil {
 
     private final static Logger logger = Logger.getLogger(TumorTypesUtil.class);
+    private final static String TSV_HEADER = "primary\tsecondary\ttertiary\tquaternary\tquinternary\tmetamaintype\tmetacolor\tmetanci\tmetaumls";
 
     private static OncoTreeRepository oncoTreeRepository;
     @Autowired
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -88,15 +90,58 @@ public static Properties getProperties() {
...
-        Properties properties = getProperties();
-        InputStream inputStream = null;
+    public static InputStream getTumorTypeInputStream(Map<String, TumorType> tumorTypes) {
+        String tsvAsString = getTsvFromTumorTypes(tumorTypes);
         try {
-            inputStream = new FileInputStream(properties.getProperty("tumor_type_file_path"));
-        } catch (FileNotFoundException e) {
-            e.printStackTrace();
+            return new ByteArrayInputStream(tsvAsString.getBytes("UTF-8"));
+        } catch (UnsupportedEncodingException e) {
```
sheridancbio: catch block no longer needed

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -88,15 +90,58 @@ public static Properties getProperties() {
...
+        // could hard code this instead of for loop
+        for (String code : tumorTypes.keySet()) {
+            TumorType tumorType = tumorTypes.get(code);
+            // skip the root node, "TISSUE". Just add it's children
+            Map<String, TumorType> children = tumorType.getChildren();
+            for (String childCode : children.keySet()) {
+                addTumorTypeToRows(children.get(childCode), rows, new ArrayList<String>());
+            }
+        }
+        return StringUtils.join(rows, "\n");
```
sheridancbio: This does not put a newline on the last row. Should each line be terminated by a new line? I generally do (to keep all lines in the same format), but not sure what the standard is.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -88,15 +90,58 @@ public static Properties getProperties() {
...
+        List<String> row = new ArrayList<>();
+        String displayName = StringUtils.defaultString(tumorType.getName()).trim() + " (" + StringUtils.defaultString(tumorType.getCode()).trim() + ")";
+        row.addAll(parents);
+        row.add(displayName);
+
+        // Need to pad for the primary - quinternary columns if all parents are not present
+        for (int i = 4; i > parents.size(); i--) {
+            row.add("");
+        }
+
```
sheridancbio: In the future, we may have oncotrees which go deeper than 5 children. We should probably add a check in this function to see if that is the case and to throw an exception if we see a node with depth > 5.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -88,15 +90,58 @@ public static Properties getProperties() {
...
+        row.add(StringUtils.defaultString(tumorType.getMainType() != null ? tumorType.getMainType().getName() : ""));
+        row.add(StringUtils.defaultString(tumorType.getColor()));
+        row.add(StringUtils.defaultString(tumorType.getNCI()));
+        row.add(StringUtils.defaultString(tumorType.getUMLS()));
```
sheridancbio: This will eventually change to calls into oncotree-mappings api. These calls will always return null.

web/src/main/java/org/mskcc/oncotree/api/TumorTypesTxtApi.java
```diff
@@ -32,8 +36,13 @@
         produces = {"text/plain"},
         method = RequestMethod.GET)
     public ResponseEntity<InputStreamResource> tumorTypesTxtGet(
+        @ApiParam(value = "The version of tumor types. For example, 1, 1.1 Please see GitHub for released versions. ")
```
sheridancbio: This example value is not correct ... let's use oncotree_latest_stable ... or better yet, call VersionUtil.getDefaultVersion() to give an example.

web/src/main/java/org/mskcc/oncotree/api/TumorTypesTxtApi.java
```diff
@@ -32,8 +36,13 @@
...
+        @RequestParam(value = "version", required = false) String version
     ) {
-        InputStream inputStream = TumorTypesUtil.getTumorTypeInputStream();
+        Map<String, TumorType> tumorTypes = new HashMap<>();
+        Version v = (version == null) ? VersionUtil.getDefaultVersion() : VersionUtil.getVersion(version);
+        tumorTypes = CacheUtil.getOrResetTumorTypesByVersion(v);
```
sheridancbio: This will soon be eliminated (dropping use of CacheUtil .. or modifying to only cache stable taxonomies from TopBraid.)

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
zheins: We still need this catch block. The getBytes method can throw an UnsupportedEncodingException.

sheridancbio: ok

PR #52, opened 2017-06-27
Add VersionUtil unit tests

sheridancbio:
- getVersion()
- getVersion(String)
- getDefaultVersion()

### DIFF
--- a/core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
+++ b/core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
@@ -1,0 +1,227 @@
+/** Copyright (c) 2017 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.  The software and
+ * documentation provided hereunder is on an "as is" basis, and
+ * Memorial Sloan-Kettering Cancer Center
+ * has no obligations to provide maintenance, support,
+ * updates, enhancements or modifications.  In no event shall
+ * Memorial Sloan-Kettering Cancer Center
+ * be liable to any party for direct, indirect, special,
+ * incidental or consequential damages, including lost profits, arising
+ * out of the use of this software and its documentation, even if
+ * Memorial Sloan-Kettering Cancer Center
+ * has been advised of the possibility of such damage.
+*/
+
+package org.mskcc.oncotree.utils;
+
+import java.io.*;
+import java.nio.file.*;
+import java.util.*;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.runner.RunWith;
+import org.junit.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.mskcc.oncotree.error.InvalidVersionException;
+import org.mskcc.oncotree.model.MainType;
+import org.mskcc.oncotree.model.Version;
+import org.mskcc.oncotree.topbraid.OncoTreeNode;
+import org.mskcc.oncotree.topbraid.OncoTreeVersionRepository;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.test.context.web.WebAppConfiguration;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.*;
+
+
+@RunWith(SpringRunner.class)
+@Import(VersionUtilTestConfig.class)
+public class VersionUtilTest {
+    private List<Version> oncoTreeVersionRepositoryMockResponse = null;
+    @Autowired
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+            fail("returned value from getDefaultVersion() was empty");
+        }
+        if (!returnedVersion.getVersion().equals(VersionUtil.DEFAULT_VERSION)) {
+            fail("returned value from getDefaultVersion() did have the correct version identifier");
+        }
+    }
+
+    private static String[] getRawTestValueSource() {
+        String[] rawTestValueSource = {
+                "oncotree_latest_stable", "urn:x-evn-master:oncotree_2017_06_21", "This is an alias for whatever OncoTree version is the latest stable (timestamped) release.",
```
zheins: Should this maybe get modeled instead of having arrays of strings that have no clear start/end for a specific version? The parsing code is also a little confusing at first glance.

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+            return value2 == null || value2.trim().length() == 0;
+        }
+        return value1.equals(value2);
+    }
+
+    @Test
+    public void testFullVersionList() throws Exception {
+        setupForTest();
+        List<Version> versionsFromUtil = versionUtil.getVersions();
+        HashMap<String, Version> returnedVersionMap = new HashMap<>(versionsFromUtil.size());
```
zheins: Use generic `Map` here

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+        for (Version version : versionsFromUtil) {
+            returnedVersionMap.put(version.getVersion(), version);
+        }
+        int failureCount = 0;
+        StringBuilder failureReport  = new StringBuilder();
+        for (String versionName : expectedVersionMap.keySet()) {
+            Version expectedVersion = expectedVersionMap.get(versionName);
+            Version returnedVersion = returnedVersionMap.get(versionName);
+            if (expectedVersion == null || returnedVersion == null) {
```
zheins: Do you need to increment `failureCount` here in this case?

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Test
+    public void testSuccessfulVersionQuery() throws Exception {
+        setupForTest();
+        String testVersionName = expectedVersionMap.keySet().iterator().next();
+        Version expectedVersion = expectedVersionMap.get(testVersionName);
+        Version returnedVersion = versionUtil.getVersion(testVersionName);
+        int failureCount = 0;
+        StringBuilder failureReport  = new StringBuilder();
+        if (expectedVersion == null || returnedVersion == null) {
```
zheins: Failure count should be incremented here

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+            if (expectedVersion == null) {
+                failureReport.append(testVersionName + " : no values in expected version hash (internal error)\n");
+            }
+            if (returnedVersion == null) {
+                failureReport.append(testVersionName + " : version expected but not returned from versionUtil\n");
+            }
+        } else {
+            String expectedVersionName = expectedVersion.getVersion();
```
zheins: Is all this code exactly the same as above? If so maybe it can be it's own method to minimize code duplication.

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+            fail(Integer.toString(failureCount) + " failed test conditions. Details follow... " + failureReport.toString());
+        }
+        return;
+    }
+
+    @Test(expected = InvalidVersionException.class)
+    public void testUnmatchedVersionQuery() throws Exception {
+        setupForTest();
+        String testVersionName = "";
+        for (String versionName : expectedVersionMap.keySet()) {
```
zheins: So this is just appending all the versions together into one long string? Is this better than just hardcoding some value you know will never be a version name?

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -0,0 +1,271 @@
...
+                Version nextVersion = new Version();
+                nextVersion.setVersion(rawTestValueSource[pos]);
+                nextVersion.setGraphURI(rawTestValueSource[pos + 1]);
+                nextVersion.setDescription(rawTestValueSource[pos + 2]);
+                expectedVersionMap.put(rawTestValueSource[pos], nextVersion);
+            }
+        }
+    }
+
+    private void setupForTest() throws Exception {
```
sheridancbio: fixed

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: fixed

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: yes, fixed

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: fixed

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: The second code block has no loop since there is only one case. But I've extract the shared inner block.

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: fixed

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
sheridancbio: fixed. Now modeled without array of raw values.

PR #54, opened 2017-06-27
Drop mainTypes api endpoints and util

sheridancbio:
- leave id field in MainType model null for now (to keep schema consistent)


zhx828:
I think there is a miscommunication here. The endpoint `/mainTypes` is needed for OncoKB.

sheridancbio:
Thanks for clarifying @zhx828 ... we will implement this endpoint properly and not delete it.

PR #56, opened 2017-06-28
Connect mainTypes endpoint to topbraid

zheins:
Connects the mainTypes endpoint to topbraid using the tumorTypesUtil class. MainTypes are populated from the same oncotree response by just going through all nodes and adding them to a set of MainTypes.

Confirmed from @zhx828 that the /api/mainTypes/{id} endpoint is not needed.

Signed-off-by: zheins <zackheins@gmail.com>

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/utils/MainTypesUtil.java
+++ b/core/src/main/java/org/mskcc/oncotree/utils/MainTypesUtil.java
@@ -1,35 +1,43 @@
 package org.mskcc.oncotree.utils;
 
+import com.google.common.collect.Lists;
 import org.mskcc.oncotree.error.InvalidOncoTreeDataException;
 import org.mskcc.oncotree.model.MainType;
 import org.mskcc.oncotree.model.Version;
 
-import java.util.List;
+import java.util.*;
+import org.mskcc.oncotree.model.TumorType;
 
 /**
  * Created by Hongxin on 2/25/16.
  */
 public class MainTypesUtil {
-    /**
-     * return MainType if not exist, create a new one.
-     *
-     * @param name The searched main type name
-     * @return
-     */
-    public static MainType getOrCreateMainType(String name, Version version) throws InvalidOncoTreeDataException {
-        List<MainType> mainTypes = CacheUtil.getMainTypesByVersion(version);
 
-        for (MainType mainType : mainTypes) {
-            if (mainType.getName().equalsIgnoreCase(name)) {
-                return mainType;
+    public static List<MainType> getMainTypesByTumorTypes(Map<String, TumorType> tumorTypes) {
+        Set<MainType> mainTypes = new HashSet<>();
+        // skip the root node, "TISSUE". Just add it's children
+        for (String tumorTypeName : tumorTypes.keySet()) {
+            TumorType tumorType = tumorTypes.get(tumorTypeName);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mskcc/oncotree/utils/MainTypesUtil.java
```diff
@@ -1,35 +1,43 @@
...
+        Set<MainType> mainTypes = new HashSet<>();
+        // skip the root node, "TISSUE". Just add it's children
+        for (String tumorTypeName : tumorTypes.keySet()) {
+            TumorType tumorType = tumorTypes.get(tumorTypeName);
+            Map<String, TumorType> children = tumorType.getChildren();
+            for (String code : children.keySet()) {
+                addMainTypesToSet(children.get(code), mainTypes);
             }
         }
+        List<MainType> toReturn = new ArrayList<>();
```
sheridancbio: ArrayList has a constructor where you can pass in a collection ... rather than doing this in two steps.
Lines 26-28 probably can be "return new ArrayList<>(mainTypes)"

web/src/main/java/org/mskcc/oncotree/api/MainTypesApi.java
```diff
@@ -34,7 +37,7 @@
         produces = {"application/json"},
         method = RequestMethod.GET)
     public ResponseEntity<MainTypesResp> mainTypesGet(
-        @ApiParam(value = "The version of tumor types. For example, 1, 1.1 Please see GitHub for released versions.")
+        @ApiParam(value = "The version of tumor types. For example, " + VersionUtil.DEFAULT_VERSION  + ". Please see the versions api documentation for released versions.")
```
sheridancbio: good

web/src/main/java/org/mskcc/oncotree/api/MainTypesApi.java
```diff
@@ -46,49 +49,11 @@
         meta.setCode(200);
         resp.setMeta(meta);
 
+        Map<String, TumorType> tumorTypes = new HashMap<>();
         Version v = (version == null) ? VersionUtil.getDefaultVersion() : VersionUtil.getVersion(version);
-        resp.setData(CacheUtil.getMainTypesByVersion(v));
+        tumorTypes = CacheUtil.getOrResetTumorTypesByVersion(v);
```
sheridancbio: We are missing a service layer, so some code like this sits here. In general, I think all data logic (aside from simple formatting/mapping) should move (in the future) out of the controller.

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -72,7 +72,7 @@
         produces = {"application/json"},
         method = RequestMethod.GET)
     public ResponseEntity<InlineResponse200> tumorTypesGet(
-        @ApiParam(value = "The version of tumor types. For example, 1, 1.1 Please see GitHub for released versions. ")
+        @ApiParam(value = "The version of tumor types. For example, " + VersionUtil.DEFAULT_VERSION + ". Please see the versions api documentation for released versions.")
```
sheridancbio: good

sheridancbio:
merging now, to be able to write unit tests

PR #58, opened 2017-06-29
Change how versions are displayed

zheins:
Instead of a span showing the currently displayed tree with the selector underneath to select a different one, put them together on one line. The selector show the currently selected tree always.

@sheridancbio

### DIFF
--- a/web/src/main/resources/static/js/main.js
+++ b/web/src/main/resources/static/js/main.js
@@ -78,12 +78,8 @@ $(document).ready(function(){
     }
 
     function initVersionsLink() {
-      // Update other versions info
-      var _str = [];
-      var _version = version_ === '' ? 'realtime' : version_;
-      Object.keys(versions_).sort().filter(function(item) {
-        return item !== _version;
-      }).forEach(function(item) {
+      var _str = [];      
+      Object.keys(versions_).sort().forEach(function(item) {
         var _hash = '#/home';
         var _content = 'Lastest';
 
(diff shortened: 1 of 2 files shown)


web/src/main/resources/static/js/main.js
```diff
@@ -78,20 +78,24 @@ $(document).ready(function(){
...
 
     function initVersionsLink() {
-      // Update other versions info
-      var _str = [];
-      var _version = version_ === '' ? 'realtime' : version_;
-      Object.keys(versions_).sort().filter(function(item) {
-        return item !== _version;
-      }).forEach(function(item) {
+      var _str = [];      
+      Object.keys(versions_).sort().forEach(function(item) {
```
sheridancbio: with this change we are displaying all versions

web/src/main/resources/static/js/main.js
```diff
@@ -78,20 +78,24 @@ $(document).ready(function(){
...
         var _hash = '#/home';
         var _content = 'Lastest';
 
         if (item !== 'realtime') {
           _hash += '?version=' + item;
           _content = item;
         }
-        _str.push('<option title="' + versions_[item].description + '" ' + 'class="item" hash="' + _hash + '">' + _content + '</option>');
+        var option = '<option title="' + versions_[item].description + '" ' + 'class="item" hash="' + _hash + '">' + _content + '</option>';
```
sheridancbio: This logic moves the current version selected into the first position in the list.
I'm guessing that this is so that the current version is displayed/visible in the select control ... but I think that doing it this way might be confusing, since it means that the list is being reordered during each selection. If a user wanted to cycle through all versions they (probably) will move from the first choice in the dropdown to the second position, third position, fourth position. But since the list order changes after each selection, this will not work as expected.
If it would work to mark the currently selected item in the (fixed order) list instead, I think that would be better. Using the 'checked' attribute in the html tag also (according to documentation) moves the item to the front of the list, so that is not a fix. Instead, I think the list could be populated in the natural order (without item swapping),  then a line could be added after line 100 (the line with .html(_str.join('')) which used jquery to adjust the property of which item is currently selected. see https://api.jquery.com/prop/ ... maybe something like https://stackoverflow.com/questions/4864620/use-jquery-to-select-a-dropdown-option if the index of the item to be selected is tracked, or by directly selecting the option with jquery if a distinguishing element id is added to the option which is desired to be the selected item.
It is possible that this initialization of the actually selected item would need to be done in a $document.ready() handler ... so that the selection occurs after the DOM is loaded.

PR #59, opened 2017-06-29
Adding unit tests for MainTypesUtil

zheins:
Also did some refactoring to put code shared between tests into a common spring configuration.
Chanted the mainTypesUtil and mainTypes api to use a flattened version of the tumorTypes map (don't need the tree structure for main types).

Signed-off-by: zheins <zackheins@gmail.com>


core/src/test/java/org/mskcc/oncotree/utils/MainTypesUtilTest.java
```diff
@@ -0,0 +1,84 @@
...
+                failureReport.append("MainType " + mainType.getName() + " from returned MainType list not found in expected MainType list\n");
+            }
+        }
+        if (failureCount > 0) {
+            fail(Integer.toString(failureCount) + " failed test conditions. Details follow...\n" + failureReport.toString());
+        }
+    }
+    
+    private Set<TumorType> getTumorTypeSet(Map<String, TumorType> expectedTumorTypeMap) {
+        Set<TumorType> tumorTypeSet = new HashSet<>();
```
sheridancbio: I think this function could be reduced to:
 return new HashSet<TumorType>(expectedTumorTypeMap.values());

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -0,0 +1,826 @@
...
+    public TumorTypesUtil tumorTypesUtil() {
+        return new TumorTypesUtil();
+    }
+    
+    @Bean
+    public MainTypesUtil mainTypesUtil() {
+        return new MainTypesUtil();
+    }
+    
+    private Version setupMockVersion() {
```
sheridancbio: note to self: we may need more of these later for testing variation across different versions

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -0,0 +1,826 @@
...
+            throw new Exception("Error : malformed rawTestValueSource");
+        }
+        final int caseCount = rawTestValueSource.length / valuesPerCase;
+        if (caseCount < 1) {
+            throw new Exception("Error : no test cases defined in rawTestValueSource");
+        }
+        for (int pos = 0; pos < rawTestValueSource.length; pos = pos + valuesPerCase) {
+            MainType nextType = new MainType();
+            String mainType = rawTestValueSource[pos + 2];
+            if (mainType != null && mainType.trim().length() > 0) {
```
sheridancbio: also .. add a check here for rawTestValueSource[pos + 4] != null ... to skip the "TISSUE" node (and other top level nodes if we add detached cell line trees later)

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -0,0 +1,826 @@
...
+                nextType.setName(mainType);
+            }
+            expectedMainTypeSet.add(nextType);
+        }
+        expectedMainTypeList.addAll(expectedMainTypeSet);
+        return expectedMainTypeList;
+    }
+
+    private static String[] getRawTestValueSource() {
```
sheridancbio: I made this, and it is ugly. There is probably a better way to import mock object data with the spring framework. If possible, I'll try to figure this out in a future PR

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -172,7 +72,6 @@ private boolean testValuesMatch(String value1, String value2) {
 
     @Test
     public void testFullOncoTreeCodeListByCode() throws Exception {
-        setupForTest();
```
sheridancbio: good .. @Before handles this

web/src/main/java/org/mskcc/oncotree/api/MainTypesApi.java
```diff
@@ -1,10 +1,27 @@
...
+*/
+
 package org.mskcc.oncotree.api;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponses;
-import java.util.HashMap;
+import java.util.*;
```
sheridancbio: If we use wildcard, we can drop line 25 too.

core/src/test/java/org/mskcc/oncotree/utils/MainTypesUtilTest.java
zheins: doh. Great point.

core/src/main/java/org/mskcc/oncotree/utils/MainTypesUtil.java
```diff
@@ -13,31 +30,16 @@
...
         Set<MainType> mainTypes = new HashSet<>();
         // skip the root node, "TISSUE". Just add it's children
-        for (String tumorTypeName : tumorTypes.keySet()) {
-            TumorType tumorType = tumorTypes.get(tumorTypeName);
-            Map<String, TumorType> children = tumorType.getChildren();
-            for (String code : children.keySet()) {
-                addMainTypesToSet(children.get(code), mainTypes);
+        for (TumorType tumorType : tumorTypes) {
+            if (tumorType.getMainType() != null && tumorType.getParent() != null) {
+                mainTypes.add(tumorType.getMainType());
```
zheins: Right, much cleaner.

core/src/main/java/org/mskcc/oncotree/utils/MainTypesUtil.java
```diff
@@ -13,31 +30,14 @@
...
-        if (children.size() > 0) {
-            for (String code : children.keySet()) {
-                TumorType child = children.get(code);
-                addMainTypesToSet(child, mainTypes);
+        for (TumorType tumorType : tumorTypes) {
+            if (tumorType.getMainType() != null && tumorType.getParent() != null) {
+                mainTypes.add(tumorType.getMainType());
             }
         }
+        return new ArrayList<>(mainTypes);
```
sheridancbio: oh .. it gets the type parameter from the function return type (diamond operator) .. cool.

PR #60, opened 2017-07-03
Corrections to prevent errors and warnings

sheridancbio:
- OncotreeUtilTestConfig Beans return concrete types (rather than interfaces) to avoid Bean selection problems on some platforms (linux)
- TopBraidRepository uses annonymous type parameter specifier to avoid unchecked type conversion warnings

### DIFF
--- a/core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
+++ b/core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
@@ -18,7 +18,6 @@
 package org.mskcc.oncotree.utils;
 
 import java.util.*;
-import static org.junit.Assert.fail;
 import org.mockito.Mockito;
 import org.mskcc.oncotree.model.MainType;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -147,13 +145,12 @@ private Version setupMockVersion() {
...
+        return expectedTumorTypeMap;
     }
-    
-    
-    private List<MainType> setupExpectedMainTypeList() throws Exception {
+
+    private ArrayList<MainType> setupExpectedMainTypeList() throws Exception {
         Set<MainType> expectedMainTypeSet = new HashSet<>();
-        List<MainType> expectedMainTypeList = new ArrayList<>();
+        ArrayList<MainType> expectedMainTypeList = new ArrayList<>();
```
zheins: Why is this necessary?

sheridancbio:
Improvement and reversion to generic collection interface types planned after release.

PR #61, opened 2017-07-03
Unit tests tumortypes txt

sheridancbio:
Add unit test for TumorTypes_txt util
Also add MainType tests for TumorTypes util

### DIFF
--- a/core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
+++ b/core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
@@ -18,7 +18,6 @@
 package org.mskcc.oncotree.utils;
 
 import java.util.*;
-import static org.junit.Assert.fail;
 import org.mockito.Mockito;
 import org.mskcc.oncotree.model.MainType;
 import org.mskcc.oncotree.model.TumorType;
@@ -37,48 +36,48 @@ import static org.junit.Assert.fail;
 @Configuration
 @ComponentScan(basePackages = {"org.mskcc.oncotree.utils","org.mskcc.oncotree.topbraid"})
 public class OncotreeUtilTestConfig {
-    
+
     @Bean
     public ArrayList<OncoTreeNode> oncoTreeRepositoryMockResponse() throws Exception {
         return setupOncotreeRepositoryMockResponse();
     }
-    
+
     @Bean
-    public Map<String, TumorType> expectedTumorTypeMap() throws Exception {
+    public HashMap<String, TumorType> expectedTumorTypeMap() throws Exception {
         return setupExpectedTumorTypeMap();
     }
-    
+
     @Bean
-    public List<MainType> expectedMainTypeList() throws Exception {
+    public ArrayList<MainType> expectedMainTypeList() throws Exception {
         return setupExpectedMainTypeList();
     }
-    
+
     @Bean
     public Version mockVersion() {
         return setupMockVersion();
     }
-    
+
     @Bean
     public OncoTreeRepository oncoTreeRepository() {
         return Mockito.mock(OncoTreeRepository.class);
     }
-    
+
     @Bean
     public TumorTypesUtil tumorTypesUtil() {
         return new TumorTypesUtil();
     }
-    
+
     @Bean
     public MainTypesUtil mainTypesUtil() {
         return new MainTypesUtil();
     }
-    
+
     private Version setupMockVersion() {
         Version mockVersion = new Version();
         mockVersion = new Version();
         mockVersion.setVersion("mockversion");
         return mockVersion;
-    }    
+    }
 
     private ArrayList<OncoTreeNode> setupOncotreeRepositoryMockResponse() throws Exception {
         ArrayList<OncoTreeNode> oncoTreeRepositoryMockResponse = new ArrayList<>();
@@ -106,9 +105,8 @@ public class OncotreeUtilTestConfig {
         }
         return oncoTreeRepositoryMockResponse;
     }
-    
-    private Map<String, TumorType> setupExpectedTumorTypeMap() throws Exception {
-        Map<String, TumorType> expectedTumorTypeMap = new HashMap<>();
+
+    private HashMap<String, TumorType> setupExpectedTumorTypeMap() throws Exception {
         String[] rawTestValueSource = getRawTestValueSource();
         final int valuesPerCase = 5;
         if (rawTestValueSource.length % valuesPerCase != 0) {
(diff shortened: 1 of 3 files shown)


sheridancbio:
This PR is not yet complete .. and in any case should not be merged before #60 (which is already incorporated into this branch).

To complete this PR, the children attributes will be set in the construction of the expectedTumorTypes from the TestConfig, so that they can be tested directly in the TumorTypes test, and also so that they can be used during the construction of the InputStream during testing of the tumorTypes_txt util.

sheridancbio:
This PR now can be merged once #60 has been approved/merged. The unit test can be improved -- for now it is testing that the output TSV spreadsheet has the correct number of columns, contains the correct header, and has the correct number of data lines based on the (mocked) oncoTree full repository.

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -37,48 +36,48 @@
...
+
     @Bean
     public ArrayList<OncoTreeNode> oncoTreeRepositoryMockResponse() throws Exception {
         return setupOncotreeRepositoryMockResponse();
     }
-    
+
     @Bean
-    public Map<String, TumorType> expectedTumorTypeMap() throws Exception {
+    public HashMap<String, TumorType> expectedTumorTypeMap() throws Exception {
```
zheins: Why is this necessary?

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -37,48 +36,48 @@
...
         return setupExpectedTumorTypeMap();
     }
-    
+
     @Bean
-    public List<MainType> expectedMainTypeList() throws Exception {
+    public ArrayList<MainType> expectedMainTypeList() throws Exception {
```
zheins: Why is this necessary?

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -118,7 +116,7 @@ private Version setupMockVersion() {
         if (caseCount < 1) {
             throw new Exception("Error : no test cases defined in rawTestValueSource");
         }
-        expectedTumorTypeMap = new HashMap<>(caseCount + 1);
+        HashMap<String, TumorType> expectedTumorTypeMap = new HashMap<>(caseCount + 1);
```
zheins: Why is this necessary?

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -147,13 +145,12 @@ private Version setupMockVersion() {
...
             expectedTumorTypeMap.put(code, nextType);
         }
-        return expectedTumorTypeMap;        
+        return expectedTumorTypeMap;
     }
-    
-    
-    private List<MainType> setupExpectedMainTypeList() throws Exception {
+
+    private ArrayList<MainType> setupExpectedMainTypeList() throws Exception {
```
zheins: Why is this necessary?

core/src/test/java/org/mskcc/oncotree/utils/OncotreeUtilTestConfig.java
```diff
@@ -147,13 +145,12 @@ private Version setupMockVersion() {
...
         Set<MainType> expectedMainTypeSet = new HashSet<>();
-        List<MainType> expectedMainTypeList = new ArrayList<>();
+        ArrayList<MainType> expectedMainTypeList = new ArrayList<>();
```
zheins: Why is this necessary?

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -45,24 +47,31 @@
...
-    public Set<TumorType> simulateTumorTypeApiResponse() throws Exception {
+    public Set<TumorType> simulateFullTumorTypeApiResponse() throws Exception {
         Map<String, TumorType> tumorTypeMap = tumorTypesUtil.getTumorTypesByVersionFromRaw(mockVersion);
         Set<TumorType> tumorTypeSet = tumorTypesUtil.flattenTumorTypes(tumorTypeMap, null);
         return tumorTypeSet;
     }
 
+    private Map<String, TumorType> getReturnedFullTumorTypeMap() throws Exception {
+        Set<TumorType> tumorTypesAsPreparedForTumorTypesEndpoint = simulateFullTumorTypeApiResponse();
+        HashMap<String, TumorType> returnedTumorTypeMap = new HashMap<>(tumorTypesAsPreparedForTumorTypesEndpoint.size());
```
zheins: Why does this need to be initialized specifically as `HashMap` and not `Map`?

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -72,22 +81,20 @@ private boolean testValuesMatch(String value1, String value2) {
...
     @Test
     public void testFullOncoTreeCodeListByCode() throws Exception {
-        Set<TumorType> tumorTypesAsPreparedForTumorTypesEndpoint = simulateTumorTypeApiResponse();
-        HashMap<String,TumorType> returnedTumorTypeMap = new HashMap<>(tumorTypesAsPreparedForTumorTypesEndpoint.size());
-        for (TumorType tumorType : tumorTypesAsPreparedForTumorTypesEndpoint) {
-            returnedTumorTypeMap.put(tumorType.getCode(), tumorType);
-        }
         int failureCount = 0;
         StringBuilder failureReport  = new StringBuilder();
+        Map<String,TumorType> returnedTumorTypeMap = getReturnedFullTumorTypeMap();
```
zheins: This one is not a `HashMap` specifically. Is this because it is not what is returned?

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -102,6 +109,29 @@ public void testFullOncoTreeCodeListByCode() throws Exception {
                     failureReport.append(makeMismatchMessage(oncoTreeCode, "name", returnedName, expectedName));
                     failureCount = failureCount + 1;
                 }
+                MainType expectedMainType = expectedTumorType.getMainType();
+                MainType returnedMainType = returnedTumorType.getMainType();
+                if (expectedMainType == null || returnedMainType == null) {
```
zheins: Do you care if they are both null? I guess do you need this outer if statement here? That last else statement could just go after these two checks if one is null and not the other. Not sure if it matters though, performance could be better this way if we don't expect it to happen often (test passes).

sheridancbio: If either is null, then a null pointer exception could thrown on line 131 when the .getName() method is called

PR #62, opened 2017-07-05
Add additional basic checks for the tumortypes_txt api util

sheridancbio:
- header is present?
- column count is correct for all rows?
- correct number of rows are present based on size of the mocked oncotree repository (full version)


sheridancbio:
changes made to revert to using generic java container interface for bean object types in unit tests.
added "Resource" annotation in place of Autowire for data objects wired by spring.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -49,7 +49,7 @@
 public class TumorTypesUtil {
 
     private final static Logger logger = Logger.getLogger(TumorTypesUtil.class);
-    private final static String TSV_HEADER = "primary\tsecondary\ttertiary\tquaternary\tquinternary\tmetamaintype\tmetacolor\tmetanci\tmetaumls";
+    public final static String TSV_HEADER = "primary\tsecondary\ttertiary\tquaternary\tquinternary\tmetamaintype\tmetacolor\tmetanci\tmetaumls";
```
zheins: Public now as it is used in a test

core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -155,12 +156,58 @@ public void testFullOncoTreeCodeListByCode() throws Exception {
...
+                        " seen: " + Integer.toString(lineColumnCount) + " on line:\n\t:" + nextLine);
+                failureCount = failureCount + 1;
+            }
+        }
+        if (expectedDataLineCount != linesRead) {
+            failureReport.append("Incorrect number of data lines. Expected: " + Integer.toString(expectedDataLineCount) + " received: " + Integer.toString(linesRead));
+            failureCount = failureCount + 1;
+        }
+        //TODO: add test for alphabetized order and the proper filling in of parents in primary, secondary, tertiary...; also meta fields
         if (failureCount > 0) {
```
zheins: Just had a thought about this pattern thats being used throughout the oncotree tests: should we have multiple checks throughout a single test that append to a stringbuilder/counter, or have all these as separate tests that get called individually? You could have all the string parsing and stuff in some setup `@Before` method and the tests just operate on those. Not sure what the right answer is - what is currently being done is fine to me. @angelicaochoa, @n1zea144  maybe you can comment?

sheridancbio: I have added comments in the code like this:
//TODO: move towards a more fine-grained test definition --- each condition being tested should be a single test function, and needed expected and actual data structures should be set up using the Before or BeforeClass annotation

sheridancbio:
Added a basic favicon.ico file (to reduce errors in logs)

PR #73, opened 2017-08-09
Retrieving the NCI and UMLS codes from crosswalk

mandawilson:
An oncotree code can have more than one NCI code and more than one UMLS code, so all codes are displayed in front end and returned by API.

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
+++ b/core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2017 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE.  The software and
+ * documentation provided hereunder is on an "as is" basis, and
+ * Memorial Sloan-Kettering Cancer Center
+ * has no obligations to provide maintenance, support,
+ * updates, enhancements or modifications.  In no event shall
+ * Memorial Sloan-Kettering Cancer Center
+ * be liable to any party for direct, indirect, special,
+ * incidental or consequential damages, including lost profits, arising
+ * out of the use of this software and its documentation, even if
+ * Memorial Sloan-Kettering Cancer Center
+ * has been advised of the possibility of such damage.
+*/
+
+package org.mskcc.oncotree.crosswalk;
+
+import java.util.*;
+import javax.annotation.PostConstruct;
+
+import org.apache.log4j.Logger;
+import org.mskcc.oncotree.topbraid.OncoTreeNode;
+import org.mskcc.oncotree.topbraid.OncoTreeRepository;
+import org.mskcc.oncotree.utils.VersionUtil;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.EnableScheduling;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Component;
+
+
+/**
+ *
+ * @author Manda Wilson
+ **/
+@Component
+@EnableScheduling
+public class MSKConceptCache {
+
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -0,0 +1,87 @@
...
+        if (oncoTreeCodesToMSKConcepts.containsKey(oncoTreeCode)) {
+            logger.debug("get(" + oncoTreeCode + ") -- in cache");
+            return oncoTreeCodesToMSKConcepts.get(oncoTreeCode);
+        }
+        logger.debug("get(" + oncoTreeCode + ") -- NOT in cache, query crosswalk");
+        return getFromCrosswalkAndSave(oncoTreeCode);
+    }
+
+    @PostConstruct // call when constructed
+    @Scheduled(cron="0 0 3 * * SUN") // call every Sunday at 3am
```
mandawilson: Yes I was excited to find that!

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -0,0 +1,87 @@
...
+    private void resetCache() {
```
mandawilson: You can define when cache elements (individually) should expire and I think it is pretty flexible, and in fact you can just have Spring handle all the caching for you (instead of creating a cache like I did), but I thought this might be better since we know that most will be loaded together all at once, so expiring individual ones might not really be useful.  Probably almost all would expire at once anyway.  We might learn more about what makes sense over time.

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -16,8 +17,8 @@
     private String color = null;
     private String name = null;
     private MainType mainType = null;
-    private String NCI = null;
-    private String UMLS = null;
+    private List<String> NCI = new ArrayList<String>();
```
mandawilson: Yeah, I guess it was change the API or randomly pick an NCI/UMLS code.

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -262,8 +269,8 @@ public String toString() {
         sb.append("  color: ").append(color).append("\n");
         sb.append("  name: ").append(name).append("\n");
         sb.append("  mainType: ").append(mainType).append("\n");
-        sb.append("  NCI: ").append(NCI).append("\n");
-        sb.append("  UMLS: ").append(UMLS).append("\n");
+        sb.append("  NCI: ").append(StringUtils.join(NCI, ",")).append("\n");
```
mandawilson: Nope you were not wrong.  The API outputs this now: 
"NCI": [
                "C3246",
                "C7052"
              ],
              "UMLS": [
                "C0026948",
                "C1266121"
              ]

core/src/main/java/org/mskcc/oncotree/utils/ListUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.*;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ *
+ * @author Manda Wilson
+ **/
+public class ListUtil { 
+
+    public static boolean containsIgnoreCase(List<String> list, String soughtFor) {
```
mandawilson: Yes definitely could be reworded!

web/src/main/resources/static/js/tree.js
```diff
@@ -339,9 +363,14 @@ var tree = (function () {
                     '<br/>';
                 _qtipContent += '<b>Name:</b> ' + d.name.replace(/\(\w+\)/gi, '') + '<br/>';
                 _qtipContent += '<b>Main type:</b> ' + d.mainType + '<br/>';
-// Temporarily removed until we properly fetch nci/umls values
-//                _qtipContent += '<b>NCI:</b> ' + nci_link  + '<br/>';
-//                _qtipContent += '<b>UMLS:</b> ' + umls_link  + '<br/>';
+                _qtipContent += '<b>NCI:</b> ' + nci_links.join(",") + '<br/>';
+                _qtipContent += '<b>UMLS:</b> ' + umls_links.join(",") + '<br/>';
+                //$.each(nci_links, function ( index, value ) {
```
mandawilson: The commented out code should definitely not be there.  That was definitely a mistake.

PR #78, opened 2017-08-29
Store and display/return previous oncotree codes

mandawilson:


core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConcept.java
```diff
@@ -37,6 +37,7 @@
     private HashMap<String, List<String>> crosswalks;
     @JsonIgnore
     private Map<String, Object> additionalProperties = new HashMap<String, Object>();
+    private Set<String> history = new HashSet<String>();
```
sheridancbio: Now each MSKConcept object has a history (previous_code) element

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConcept.java
```diff
@@ -81,6 +82,30 @@ public void setCrosswalks(HashMap<String, List<String>> crosswalks) {
...
+    */
+    public Set<String> getHistory() {
+        return history;
+    }
+
+    /**
+    *
+    * @param oncotreeCode
+    */
+    public void addHistory(String oncotreeCode) {
```
sheridancbio: A little unusual to have a get() / add() combination, but it seems to work here. To be truly multi-purpose, a set() method could be provided which overwrote the set, but we don't need it now.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -41,6 +42,8 @@
 
     private final static Logger logger = Logger.getLogger(MSKConceptCache.class);
     private static HashMap<String, MSKConcept> oncoTreeCodesToMSKConcepts = new HashMap<String, MSKConcept>();
+    // use this to store and look up previous oncoTree codes
+    private static HashMap<String, HashSet<String>> topBraidURIsToOncotreeCodes = new HashMap<String, HashSet<String>>();
```
sheridancbio: a global cache is added here to hold all mappings of oncotree concept URIs (topbraid Identifiers) to the oncotree codes they might refer to.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -67,20 +70,38 @@ public MSKConcept get(String oncoTreeCode) {
     private void resetCache() {
         logger.info("resetCache() -- clearing Crosswalk MSKConcept cache and refilling");
         oncoTreeCodesToMSKConcepts.clear();
-        List<OncoTreeNode> oncoTreeNodes = oncoTreeRepository.getOncoTree(VersionUtil.getDefaultVersion());
-        for (OncoTreeNode node : oncoTreeNodes) {
-            getFromCrosswalkAndSave(node.getCode());
-        } 
+        // versions are ordered in ascending order by release date
+        for (Version version : VersionUtil.getVersions()) {
```
sheridancbio: Instead of default version only, the cache is loaded across all version (in forward time order)

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -67,20 +70,38 @@ public MSKConcept get(String oncoTreeCode) {
...
+            List<OncoTreeNode> oncoTreeNodes = oncoTreeRepository.getOncoTree(version);
+            for (OncoTreeNode node : oncoTreeNodes) {
+                MSKConcept mskConcept = getFromCrosswalkAndSave(node.getCode());
+                // get all codes defined so far for this topbraid uri and save in history
+                if (topBraidURIsToOncotreeCodes.containsKey(node.getURI())) {
+                    // do not add this code to the history, but add any others
+                    HashSet<String> allButThisNode = new HashSet<String>(topBraidURIsToOncotreeCodes.get(node.getURI()));
```
sheridancbio: because we are going in forward time order, this "snapshot" will show all the oncotree codes defined up to this point in time. (will only include current and previous codes)

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -67,20 +70,38 @@ public MSKConcept get(String oncoTreeCode) {
...
+                    allButThisNode.remove(node.getCode());
+                    mskConcept.addHistory(allButThisNode);
```
sheridancbio: Add the set of previous codes into the MSKConcept cache (organized by oncotree code).

So long as an oncotree code is never "resurrected", we should be ok. I think that if a code were present in version 1, renamed in version 2, renamed in version 3, and renamed back to the starting code in version 4, I think we might be in trouble. We can test this in the future.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -67,20 +70,38 @@ public MSKConcept get(String oncoTreeCode) {
...
+                } else {
+                    topBraidURIsToOncotreeCodes.put(node.getURI(), new HashSet<String>());
+                }
+                // now save this as onoctree code history for this topbraid uri
+                topBraidURIsToOncotreeCodes.get(node.getURI()).add(node.getCode());
```
sheridancbio: getURI() is a new property for this purpose

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -67,20 +70,38 @@ public MSKConcept get(String oncoTreeCode) {
...
+            }
+        }
     }
 
     private MSKConcept getFromCrosswalkAndSave(String oncoTreeCode) {
-        MSKConcept concept = null;
+        // only save if we have not seen before (UMLS/NCI info will not be different)
+        if (oncoTreeCodesToMSKConcepts.containsKey(oncoTreeCode)) {
+            return oncoTreeCodesToMSKConcepts.get(oncoTreeCode);
```
sheridancbio: ah .. maybe this covers the "returned to original code" case I mentioned above. But it also may "lose" the information from the "version 2" and "version 3" history.

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeNode.java
```diff
@@ -31,17 +31,18 @@
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonPropertyOrder({
+    "uri",
     "code",
     "name",
     "mainType",
     "color",
-    "nci",
```
sheridancbio: this is cleanup ... we don't receive these fields any more from the sparql

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeRepository.java
```diff
@@ -38,7 +38,7 @@
 
     private String query = "PREFIX skos:<http://www.w3.org/2004/02/skos/core#> " +
         "PREFIX onc:<http://data.mskcc.org/ontologies/oncotree/> " +
-        "SELECT DISTINCT ?code ?name ?mainType ?color ?nci ?umls ?icdo ?parentCode " +
+        "SELECT DISTINCT (?s AS ?uri) ?code ?name ?mainType ?color ?parentCode " +
```
sheridancbio: nice find of the uri here

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeVersionRepository.java
```diff
@@ -36,16 +36,21 @@
...
+        "SELECT ?api_identifier ?graph_uri ?description ?release_date " +
         "WHERE { " +
         "   GRAPH <urn:x-evn-master:oncotree_version> { " +
         "       ?s oncotree-version:graph_uri ?graph_uri. " +
         "       ?s oncotree-version:api_identifier ?api_identifier. " +
+        "       ?s oncotree-version:release_date ?release_date. " +
         "       OPTIONAL{?s oncotree-version:description ?description.} " +
         "   } " +
-        "}";
+        "} ORDER BY ASC(?release_date)";
```
sheridancbio: good .. let topbraid do the work

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -51,7 +51,7 @@
 public class TumorTypesUtil {
 
     private final static Logger logger = Logger.getLogger(TumorTypesUtil.class);
-    public final static String TSV_HEADER = "primary\tsecondary\ttertiary\tquaternary\tquinternary\tmetamaintype\tmetacolor\tmetanci\tmetaumls";
+    public final static String TSV_HEADER = "primary\tsecondary\ttertiary\tquaternary\tquinternary\tmetamaintype\tmetacolor\tmetanci\tmetaumls\thistory";
```
sheridancbio: tumortype.txt expansion : new column for history (maybe this should match previous_code from the tumorTypes api?

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -284,10 +285,13 @@ private static void setDepthAndTissue(TumorType tumorType, int depth, String tis
                     tumorType.setNCI(crosswalks.get("NCI"));
                 }
                 List<String> umlsIds = new ArrayList<String>();
-                for (String mskConceptId : mskConcept.getConceptIds()) {
-                    umlsIds.add(mskConceptId.replace("MSK", "C"));
+                if (mskConcept.getConceptIds() != null) {
```
sheridancbio: can we have a concept without an Id?  can't hurt to check for null I guess, but if we do have a null conceptId maybe we should log a warning here (as long as we are checking)

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -284,10 +285,13 @@ private static void setDepthAndTissue(TumorType tumorType, int depth, String tis
...
+                    for (String mskConceptId : mskConcept.getConceptIds()) {
+                        umlsIds.add(mskConceptId.replace("MSK", "C"));
+                    }
                 }
                 tumorType.setUMLS(umlsIds);
+                tumorType.setHistory(new ArrayList(mskConcept.getHistory()));
```
sheridancbio: good .. initialize list with set (collection)

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
```diff
@@ -158,9 +158,6 @@ private Version setupMockVersion() {
             nextNode.setName(rawTestValueSource[pos + 1]);
             nextNode.setMainType(rawTestValueSource[pos + 2]);
             nextNode.setColor(rawTestValueSource[pos + 3]);
-            nextNode.setNci("");
```
sheridancbio: Just a comment for future TODO : we should add a mock crosswalk repository and add unit tests.

web/src/main/resources/static/js/tree.js
```diff
@@ -113,6 +113,10 @@ var tree = (function () {
                             uniqueTreeNode[type].umls = row.metaumls;
                         }
 
+                        if(row.hasOwnProperty('history')){
+                            uniqueTreeNode[type].history = row.history;
```
sheridancbio: parse new field from tumortypes.txt output .. if we rename the field in the output of tumortypes.txt we should match it here.

web/src/main/resources/static/js/tree.js
```diff
@@ -592,7 +598,8 @@ var tree = (function () {
             if (searchKey === '') {
                 d3.select(this).style('fill', 'black');
             } else {
-                if (d.name.toLowerCase().indexOf(searchKey) !== -1) {
+                if (d.name.toLowerCase().indexOf(searchKey) !== -1 ||
+                    (typeof d.history !== 'undefined' && d.history != '' && d.history.toLowerCase().split(",").indexOf(searchKey) !== -1)) {
```
sheridancbio: very nice .. enable searching / highlighting based on the previous codes that have been loaded.
The whole search function could be made more efficient .. as a later enhancement.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
sheridancbio: oh .. it already does match .. because we are staying consistent with the current API by reusing the field "history". So never mind.

PR #82, opened 2017-12-20
Removing remote dependency

mandawilson:
Removing remote dependency on http://d3js.org/d3.v3.min.js?05022016.  Saved the file as part of the code base.  Also removed js/d3.min.js and js/d3.layout.js which look like they are no longer used.


sheridancbio:
Nice work, figuring this out so quickly

PR #84, opened 2018-01-08
Improvements for addition of oncotree-mappings

sheridancbio:
- added redirection handler for previous URL pattern (/oncotree-mappings/) to html page resource
- added javascript validation of oncotree-mappings form page, with a trap and error messages for invalid argument combinations
- added some java argument cleaning and length limiting to the oncotree-mappings web api
- argument cleaning implemented in ApiUtil component, with unit testing
- dropped non-functional web.xml file (now configured through springframework/spring-boot)
- adjusted copyright notice headers
- whitespace and style cleanup


sheridancbio:
Apologies for mixing style cleanup with code changes in the .html and the java controller.
The only semantic change in the .html file was the addition of the arguments checking and error reporting inside the final <script> element.
The controller now uses ApiUtil to clean arguments and has 4 static constants which limit the length of the cleaned arguments (overlong arguments cause an exception to be thrown before processing). The logic for checking argument validity now includes the maximum length and is refactored, but it should be conceptually the same as before.

web/src/main/java/org/mskcc/oncotree/api/OncotreeMappingsApi.java
```diff
@@ -1,97 +1,126 @@
...
+    @RequestMapping(value = "api/crosswalk", method = RequestMethod.GET)
+    public OncotreeMappingsResp getMappings(
+            @RequestParam(value="vocabularyId", required=false) String vocabularyId,
+            @RequestParam(value="conceptId", required=false) String conceptId,
+            @RequestParam(value="histologyCode", required=false) String histologyCode,
+            @RequestParam(value="siteCode", required=false) String siteCode) {
+            String cleanVocabularyId = apiUtil.cleanArgument(vocabularyId);
+            String cleanConceptId = apiUtil.cleanArgument(conceptId);
+            String cleanHistologyCode = apiUtil.cleanArgument(histologyCode);
+            String cleanSiteCode = apiUtil.cleanArgument(siteCode);
```
mandawilson: Great!

sheridancbio:
Merging now because it has been deployed. But additional reviews/feedback would be very welcome.

PR #85, opened 2018-02-01
Stop using tumor type txt

mandawilson:
The front end is now using api/tumorTypes?flat=false instead of api/tumor_types.txt to create the tree.

### DIFF
--- a/web/src/main/resources/static/js/tree.js
+++ b/web/src/main/resources/static/js/tree.js
@@ -15,11 +15,7 @@ var tree = (function () {
 
     var tree, diagonal, vis, numOfTumorTypes = 0, numOfTissues = 0;
 
-    //Hierarchical nodes.
-    var treeNode = {};
-
-    //Regardless column, each node name should be unique. Store global attributes here.
-    var uniqueTreeNode = {};
+    var oncotreeCodesToNames = {}; // used to find duplicate codes
 
     var searchResult = [];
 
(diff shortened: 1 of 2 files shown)


mandawilson:
This is a first pass at this so that you can see it.  I want to make changes, so no need to review yet.

web/src/main/resources/static/js/tree.js
```diff
@@ -39,15 +35,79 @@ var tree = (function () {
...
+    function process_children(parentNode, childData) {
+        // childData is always for a new unique node
+        var childNode = new UniqueTreeNodeDatum();
+        childNode.name = childData.name + " (" + childData.code + ")";
+        childNode.acronym = childData.code;
+
+        if (childData.hasOwnProperty('mainType') && childData.mainType != null && childData.mainType.hasOwnProperty('name') && childData.mainType.name !== '') {
+            childNode.mainType = childData.mainType.name;
+        } else {
+            childNode.mainType = 'Not Available';
```
sheridancbio: good ... this will now cover the "TISSUE" root node which is not present in tumor_types.txt

web/src/main/resources/static/js/tree.js
```diff
@@ -39,15 +35,79 @@ var tree = (function () {
...
+        parentNode.children.push(childNode);
+
+        // now process this node's children
+        if (childData.hasOwnProperty('children')) {
+            var codesSortedByName =  getOncotreeCodeKeysSortedByName(childData.children);
+            if (codesSortedByName.length > 0) {
+                childNode.children = [];
+                codesSortedByName.forEach(function (code) {
+                    // now childNode's children are added
+                    process_children(childNode, childData.children[code]);
```
sheridancbio: recursion: makes sense with our representation. If we ever create a topbraid relationship which creates a cycle in the tree we will be in trouble - but presumably it would then cease to be a "tree" and so should not happen.

web/src/main/resources/static/js/tree.js
```diff
@@ -63,67 +123,18 @@ var tree = (function () {
...
-
-                    //Ignore specific columns
-                    if (['metamaintype', 'metacolor'].indexOf(col) !== -1) break;
-
-                    if (!type) break;
-
-                    if (!(type in node)) {
-                        node[type] = {};
-                    }
+        d3.json(jsonUrl, function (oncotree_json) {
```
sheridancbio: This is awesome. Great find.

web/src/main/resources/static/js/tree.js
```diff
@@ -63,67 +123,18 @@ var tree = (function () {
...
-                        if(row.hasOwnProperty('history')){
-                            uniqueTreeNode[type].history = row.history;
-                        }
-
-                    }
-                    node = node[type];
-                }
+            getOncotreeCodeKeysSortedByName(oncotree_json.data.TISSUE.children).forEach(function (code) {
+                var childData = oncotree_json.data.TISSUE.children[code];
+                // these nodes all belong at root of tree
```
sheridancbio: I think the TISSUE node is the root of the tree, so I would say "these nodes are all children of the TISSUE root node" or "these nodes are all tissue type nodes at depth 1" or similar.

web/src/main/resources/static/js/tree.js
```diff
@@ -335,8 +321,8 @@ var tree = (function () {
                 }
 
                 var nci_links = [];
-                if (typeof d.nci !== 'undefined' && d.nci != '') {
-                    $.each(d.nci.split(","), function( index , value ) {
+                if (typeof d.nci !== 'undefined' && d.nci.length > 0) {
+                    $.each(d.nci, function( index , value ) {
```
sheridancbio: another good simplification .. not relying on the presentation style of lists from the tumort_types.txt API endpoint. This must have required some searching during testing to find cases with multiple NCI codes - nice.

PR #86, opened 2018-02-01
News tab

yichaoS:
http://dashi-dev.cbio.mskcc.org:8080/yichao-oncotree/#/home


web/src/main/resources/static/index.html
```diff
@@ -12,64 +12,138 @@
...
-            <input type="text" class="form-control" placeholder="Search Tumor Type" />
-            <span id="searchRemoveIcon" class="form-control-feedback glyphicon glyphicon-remove-circle"></span>
-            <span id="searchResult" class="form-control-feedback result"></span>
-            <span class="input-group-btn">
-              <button type="button" class="btn btn-default">Search</button>
-            </span>
 
+        <!-- news tab -->
+        <div class="tab-pane fade" id='news' role="tabpanel">
+          <div class="embed-responsive embed-responsive-16by9">
```
mandawilson: This div and iframe aren't needed since news is loaded dynamically and put inside the id='news' div.

mandawilson:
Yichao and I also talked about how pulling from an external file that contains markdown and converting it to HTML has an XSS vulnerability, described here: https://github.com/showdownjs/showdown/wiki/Markdown's-XSS-Vulnerability-(and-how-to-mitigate-it). We should all discuss this, but we do have control over the file we are pulling so I think it should be OK.

PR #90, opened 2018-02-07
Version list 

yichaoS:
- Use 'visible' attribute from AjAX result to eliminate what goes into the dropbox box
- Add a section to display description of selected version
http://dashi-dev.cbio.mskcc.org:8080/yichao-oncotree/


web/src/main/resources/static/js/main.js
```diff
@@ -20,7 +20,10 @@ $(document).ready(function(){
             && data.hasOwnProperty('data')
             && data.data instanceof Array) {
             versions_ = data.data.reduce(function(acc, cur) {
-              acc[cur.api_identifier] = cur;
+              //acc[cur.api_identifier] = cur;
```
mandawilson: I would delete this line.

web/src/main/resources/static/index.html
```diff
@@ -92,15 +88,14 @@
 
         <!-- news tab -->
         <div class="tab-pane fade" id='news' role="tabpanel">
-          <div class="embed-responsive embed-responsive-16by9">
-              <iframe class="embed-responsive-item" src="https://github.com/cBioPortal/oncotree/blob/master/docs/News.md"></iframe>
-          </div>
+          <div class="embed-responsive embed-responsive-16by9"></div>
```
mandawilson: I *think* this div can be deleted, the news is put in the id='news' div.

web/src/main/resources/static/index.html
```diff
@@ -44,16 +44,13 @@
               <br/>
               <span id="summary-info"></span>
               <br/>
-              <span>Please see <a href="swagger-ui.html" target="_blank"><b>here</b></a> for the API</span>
```
mandawilson: Good!

PR #94, opened 2018-02-08
Added new tab for oncotree-mapping and changed some style

mandawilson:
* Moved oncotree-mapping.html to tab "Mapping" on homepage
* /oncotree-mappings/ now redirects to "/#/home?tab=mapping" which will select the "Mapping" tab
* Moved JavaScript in index.html to news.js and oncotree-mapping.js
* Oncotree version dropdown no longer has tooltip with version description as the description is displayed prominently on the page.
* Experimented with style changes, to make "Home", "Mapping", and "News" tabs more uniform.  Feel free to make alternative suggestions!


web/src/main/java/org/mskcc/oncotree/OncotreeApplication.java
```diff
@@ -41,7 +41,7 @@ public void addCorsMappings(CorsRegistry registry) {
             @Override
             public void addViewControllers(final ViewControllerRegistry registry) {
                 super.addViewControllers(registry);
-                registry.addRedirectViewController("/oncotree-mappings/","/oncotree-mappings.html").setKeepQueryParams(true).setStatusCode(HttpStatus.MOVED_PERMANENTLY);
+                registry.addRedirectViewController("/oncotree-mappings/","/#/home?tab=mapping").setKeepQueryParams(true).setStatusCode(HttpStatus.MOVED_PERMANENTLY);
```
sheridancbio: We should also have a sysadmin update the apache config to forward to this new spot from the original URL directory for the prior webapp.

web/src/main/resources/static/js/news.js
```diff
@@ -0,0 +1,31 @@
+function loadNews(callback) {
```
sheridancbio: nice encapsulation of code

web/src/main/resources/static/js/oncotree-mapping.js
```diff
@@ -0,0 +1,61 @@
+$(function (){
```
sheridancbio: good code modularization

PR #97, opened 2018-03-15
Update resource_uri_to_oncocode_mapping.txt

ritikakundra:


sheridancbio:
I see that there is a new node added to oncotree_development version with Tubular Adenoma of the Colon (TAC) and URI http://data.mskcc.org/ontologies/oncotree/ONC000858

So I think the edit here was simply a typo.

PR #102, opened 2018-04-03
Code cleanup and cache refreshing

mandawilson:
* Removed meta information from JSON responses
* Removed unused code (methods, classes)
* Cleaned up code
* Cache
  - Fully initialize cache on startup
  - Refresh the cache daily at 11am
  - Added refresh endpoint
  - Expire cache if it has not been refreshed for more than 3 days
  - refreshCache endpoint hidden in Swagger

NOTE: web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java has two TODO items we should do something about.


mandawilson:
I will create a separate PR to fix the memory leak, and then will rebase this on that one.

mandawilson:
The caching code needs unit tests.

mandawilson:
@zhx828 We have made major changes here to the API, but haven't committed to them yet.  We also don't have a deployment plan yet.

mandawilson:
I will rebase once this is merged: https://github.com/cBioPortal/oncotree/pull/103

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -3,170 +3,90 @@
...
 
         // Cache in tumor types in case no data present
-        CacheUtil.getOrResetTumorTypesByVersion(v);
+        CacheUtil.getTumorTypesByVersion(v);
 
+        // each query has a list of results
         for (TumorTypeQuery query : queries.getQueries()) {
             List<TumorType> matchedTumorTypes = new ArrayList<>();
             matchedTumorTypes = v == null ? new ArrayList<TumorType>() : TumorTypesUtil.findTumorTypesByVersion(query.getType(), query.getQuery(), query.getExactMatch(), v, false);
+            // TODO add back?
```
mandawilson: We should either delete this, or add it back and fix it.

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -181,42 +101,36 @@
...
         Version v = (version == null) ? VersionUtil.getDefaultVersion() : VersionUtil.getVersion(version);
 
         // Cache in tumor types in case no data present
-        CacheUtil.getOrResetTumorTypesByVersion(v);
+        CacheUtil.getTumorTypesByVersion(v);
 
         matchedTumorTypes = v == null ? new ArrayList<TumorType>() : TumorTypesUtil.findTumorTypesByVersion(type, query, exactMatch, v, false);
-        SearchTumorTypesResp resp = new SearchTumorTypesResp();
-
+        // TODO add back or remove levels option
```
mandawilson: Also here, we should delete this or add it back and fix it.

zhx828: I do like the levels. It is quite useful when you have a nested object.

zhx828:
@mandawilson thanks for letting me know. I think it would be better to have a testing instance for all the changes, I can also do some testing with OncoKB instance connected.

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
sheridancbio: I also like the idea of specifying levels, but we should talk through the behavior and understand the design/requirements. I think that the search api endpoints always return a flat response (not nested). In this case, specifying the levels could simply act as a filter, removing all non-matching nodes from the list. If the /api/tumorTypes/search endpoint could return a nested representation (similar to the /api/tumorTypes endpoint) then I think we would have difficulty representing the child-parent relationship in the response. The field "parent" is not too much of a problem, since it is simply a string. But the field "children" is a fully populated object of type TumorType ... and if a user requested levels "2,4" then I wonder whether the level 4 nodes would appear as children of level 2? Somehow that seems incorrect. If we wanted a nested representation which restricted the levels returned we might need to change to a "filter all levels above" / "filter all levels below" approach so that there are no gap levels in between parents and children. Also, there is a question of how to contain the nodes in the uppermost (lowest numbered) level. In the GET /api/tumorTypes endpoint there is one root node (TISSUE) at level 0 which contains as children all the level 1 nodes (the tissues). If we returned a nested response from search and filtered all levels from 0 through 2 then I guess the response would be a list of nested trees, each one starting at level 3 ?

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -12,36 +29,37 @@
 @javax.annotation.Generated(value = "class io.swagger.codegen.languages.SpringMVCServerCodegen", date = "2016-04-04T17:16:11.368Z")
 public class TumorType {
 
-    private Integer id = null;
     private String code = null;
     private String color = null;
     private String name = null;
-    private MainType mainType = null;
+    private String mainType = null;
     private List<String> NCI = new ArrayList<String>();
```
sheridancbio: The fields NCI and UMLS make these links to external annotations a top-level part of the model. That is not inherently wrong, but each new mapping to an external ontology may require a change to our model and break the API for external users. (A major revision change). So to avoid that, I think we should consider representing links to external vocabularies in a single top level object ... perhaps a map from vocabulary names {NCI, UMLS, ...} to string values.

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -1,78 +1,114 @@
...
 
-    public static void resetTumorTypesByVersion(Version version) {
-        if (tumorTypes.containsKey(version)) {
-            tumorTypes.get(version).clear();
-        } else {
-            tumorTypes.put(version, new HashMap<String, TumorType>());
+    private static final Logger logger = LoggerFactory.getLogger(CacheUtil.class);
+
+    private static Map<Version, Map<String, TumorType>> tumorTypes = null;
+    private static Date lastSuccessfullRefresh = null;
```
sheridancbio: good .. a timestamp approach to knowing how stale the cache is

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -1,78 +1,114 @@
...
+    @PostConstruct // call when constructed
+    @Scheduled(cron="0 0 11 * * *") // reset at 11am every day
+    private void scheduleResetCache() {
+        // TODO make sure we don't have two scheduled calls run simultaneously
+        try {
+            resetCache();
+        } catch (TopBraidException exception) {
+            // check to see if we should expire cache 
+            ZonedDateTime now = ZonedDateTime.now();
+            ZonedDateTime threeDaysAgo = now.plusDays(-3);
```
sheridancbio: We may adjust this time period. We should probably name the variable something more abstract (cacheExpirationAge?) and have the "-3" in a final data member ... possibly in seconds to allow for more granular fine-tuning. (Or read this from the properties file)

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -1,78 +1,114 @@
...
+
+    private static Map<Version, Map<String, TumorType>> tumorTypes = null;
+    private static Date lastSuccessfullRefresh = null;
+
+    @PostConstruct // call when constructed
+    @Scheduled(cron="0 0 11 * * *") // reset at 11am every day
+    private void scheduleResetCache() {
+        // TODO make sure we don't have two scheduled calls run simultaneously
+        try {
+            resetCache();
```
sheridancbio: Here we try to reset the cache once every day at 11. This gives only 3 chances to get a successful cache reset. Alternatively we could check more frequently but only attempt a reset if the age of the current cache is over a certain age. This would perform more retries if a reset failed.
Also, once the cache has fully expired, it would be good to have another scheduled attempt to reset the cache. Maybe a function which runs every minute (or every few minutes) and only tries to reset the cache if it is currently expired.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -61,31 +56,13 @@
...
     public void setMSKConceptCache(MSKConceptCache property) { mskConceptCache = property; }
 
-    private static final String PROPERTY_FILE = "classpath:application.properties";
     private static List<String> TumorTypeKeys = Arrays.asList("code", "name", "nci", "level", "umls", "maintype", "color");
 
     public static Map<String, TumorType> getTumorTypesByVersionFromRaw(Version version) throws InvalidOncoTreeDataException {
-        Map<String, TumorType> tumorTypes = new HashMap<>();
         if (version != null) {
-            tumorTypes = loadFromRepository(version);
+            return loadFromRepository(version);
```
sheridancbio: good cleaning

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -195,7 +172,7 @@ private static void addTumorTypeToRows(TumorType tumorType, List<String> rows, L
         return tumorTypes;
     }
 
-    public static boolean hasNoParent(TumorType node) {
+    private static boolean hasNoParent(TumorType node) {
```
sheridancbio: 👍

core/src/test/java/org/mskcc/oncotree/utils/MainTypesUtilTest.java
```diff
@@ -44,32 +45,33 @@
...
 
         StringBuilder failureReport = new StringBuilder();
         int failureCount = 0;
         if (returnedMainTypeList.size() != expectedMainTypeList.size()) {
             failureCount++;
             failureReport.append("Returned MainType list is of different length than the expected MainType list." +
             "\n\tReturned list length: " + String.valueOf(returnedMainTypeList.size()) +
-            "\n\tExpected list length: " + String.valueOf(expectedMainTypeList.size()) + "\n");
+            "\n\tExpected list length: " + String.valueOf(expectedMainTypeList.size()) +
+            "\n\tDifference is: " + StringUtils.join(CollectionUtils.disjunction(returnedMainTypeList, expectedMainTypeList), ", ") + "\n");
```
sheridancbio: good output during test failure

web/src/main/java/org/mskcc/oncotree/OncotreeApplication.java
```diff
@@ -52,7 +52,7 @@ ApiInfo apiInfo() {
         ApiInfo apiInfo = new ApiInfo(
             "OncoTree API",
             "OncoTree API definition from MSKCC",
-            "0.0.1",
+            "1.0.0",
```
sheridancbio: Nice. We can start versioning our API.

web/src/main/java/org/mskcc/oncotree/api/ApiOriginFilter.java
```diff
@@ -1,27 +0,0 @@
-package org.mskcc.oncotree.api;
```
sheridancbio: We may want a filter for the resetCache endpoint (only accepting requests from inside the MSKCC network). We should talk this over.

web/src/main/java/org/mskcc/oncotree/api/MainTypesApi.java
```diff
@@ -20,57 +20,44 @@
...
-
-        Map<String, TumorType> tumorTypes = new HashMap<>();
+    ) {
         Version v = (version == null) ? VersionUtil.getDefaultVersion() : VersionUtil.getVersion(version);
-        tumorTypes = CacheUtil.getOrResetTumorTypesByVersion(v);
+        Map<String, TumorType> tumorTypes = CacheUtil.getTumorTypesByVersion(v);
         Set<TumorType> tumorTypesSet = TumorTypesUtil.flattenTumorTypes(tumorTypes, version);
-        resp.setData(MainTypesUtil.getMainTypesByTumorTypes(tumorTypesSet));
-
-        return new ResponseEntity<MainTypesResp>(resp, HttpStatus.OK);
```
sheridancbio: nice simplification through RestController

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -3,170 +3,90 @@
...
-        return new ResponseEntity<ChildrenListResp>(HttpStatus.OK);
+        Map<String, TumorType> tumorTypes = CacheUtil.getTumorTypesByVersion(v);
+        logger.debug("tumorTypesGet() -- returning " + tumorTypes.size() + " tumor types");
+        return tumorTypes;
     }
 
+    @ApiOperation(value = "Return all available tumor types as a list.", notes = "", response = TumorType.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "Tumor Types list.")})
+    @RequestMapping(value = "/flat",
```
sheridancbio: Does this mean we are dropping the parameter ?flat={true/false} ?  (splitting it into a separate endpoint)
I think that's ok, but it raises the question of the other endpoints ... should our responses to /search have a /search/flat endpoint too (to leave room for the non-flat version one level up) ?

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -3,170 +3,90 @@
...
-        @io.swagger.annotations.ApiResponse(code = 200, message = "An array of tumor types")})
+    @ApiOperation(value = "Tumor Types", notes = "...", response = TumorType.class)
+    @ApiResponses(value = {
+        @ApiResponse(code = 200, message = "An array of tumor types")})
     @RequestMapping(value = "/search",
-        produces = {"application/json"},
+        produces = {APPLICATION_JSON_VALUE},
         method = RequestMethod.POST)
-    public ResponseEntity<SearchTumorTypesPostResp> tumorTypesSearchPost(
+    public List<List<TumorType>> tumorTypesSearchPost(
```
sheridancbio: It's not immediate clear to me why this is List<List<..>>

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
sheridancbio: I think I'm in favor of the levels argument .. but see my large comment below (related to the flat/nonflat returns)

web/src/main/java/org/mskcc/oncotree/api/TumorTypesApi.java
```diff
@@ -181,42 +101,36 @@
...
+        produces = {APPLICATION_JSON_VALUE},
         method = RequestMethod.GET)
-    public ResponseEntity<SearchTumorTypesResp> tumorTypesSearchTypeQueryQueryGet(
+    public Iterable<TumorType> tumorTypesSearchTypeQueryQueryGet(
         @ApiParam(value = "Query type. It could be 'id', 'code', 'name', 'mainType', 'level', 'nci', 'umls' or 'color'. You can also use 'all' to search all content.", required = true)
         @PathVariable("type") String type,
         @ApiParam(value = "The query content", required = true)
         @PathVariable("query") String query,
-        @ApiParam(value = "The version of tumor types. For example, 1, 1.1 Please see GitHub for released versions. ")
+        @ApiParam(value = "The version of tumor types. For example, " + VersionUtil.DEFAULT_VERSION + ". Please see the versions api documentation for released versions.")
```
sheridancbio: 👍

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -12,36 +29,37 @@
...
-        this.id = id;
+    public TumorType(TumorType otherTumorType) {
+        code = otherTumorType.code;
+        color = otherTumorType.color;
+        name = otherTumorType.name;
+        mainType = otherTumorType.mainType;
+        NCI = new ArrayList<String>(otherTumorType.NCI);
+        UMLS = new ArrayList<String>(otherTumorType.UMLS);
+        tissue = otherTumorType.tissue;
+        children = new HashMap<String, TumorType>(otherTumorType.children);
```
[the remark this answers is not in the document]
mandawilson: @sheridancbio This will be replaced with https://github.com/cBioPortal/oncotree/pull/104

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -1,78 +1,114 @@
...
-    }
-
-    public static Map<String, TumorType> getTumorTypesByVersion(Version version) throws InvalidOncoTreeDataException {
         if (tumorTypes.containsKey(version)) {
-            return tumorTypes.get(version);
-        } else {
-            tumorTypes.put(version, TumorTypesUtil.getTumorTypesByVersionFromRaw(version));
-            return tumorTypes.get(version);
+            logger.debug("getTumorTypesByVersion() -- found '" + version.getVersion() + "' in cache");
+            return Collections.unmodifiableMap(tumorTypes.get(version));
```
mandawilson: This will replaced by code in https://github.com/cBioPortal/oncotree/pull/104

mandawilson:
Bump the maven version number too.

mandawilson:
Are errors returned as HTML or JSON?

sheridancbio:
I think that since we are going to deploy this code as ROOT.war in the oncotree-tomcat and then migrate to it, we can deploy now without breaking existing clients. I am removing the do-not-merge tag.

PR #103, opened 2018-04-03
Fixed memory leak

mandawilson:
The Version object is never found in the cache, and so we always get the tumor types from the repository, and then store it in the cache - so the cache was getting bigger with every single request to the website.  I did some logging and confirmed this is true, I refreshed a five of times, and ended up with five 'oncotree_development' Version objects in the cache.  It is because Version didn't have equals and hashCode overridden, and Version (instead of a String version) was the key).

Fix is here.

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
+++ b/core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
@@ -5,12 +5,18 @@ import org.mskcc.oncotree.model.MainType;
 import org.mskcc.oncotree.model.TumorType;
 import org.mskcc.oncotree.model.Version;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.*;
 
 /**
  * Created by Hongxin on 2/25/16.
  */
 public class CacheUtil {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -34,9 +40,17 @@ public static void resetTumorTypesByVersion(Version version) {
     }
 
     public static Map<String, TumorType> getTumorTypesByVersion(Version version) throws InvalidOncoTreeDataException {
+        logger.debug("getTumorTypesByVersion() -- looking for version '" + version.getVersion() + "' in cache");
```
zhx828: I never used logger before. But seems like this sentence is only needed when debug mode is enabled as well?

mandawilson: @zhx828 That line won't execute when in debug mode (because it is logger.debug) but I wrapped the for loop in the check for debug mode because otherwise the loop would execute even in non-debug mode, even if nothing got printed.  It looks weird, but I think it is OK.

PR #104, opened 2018-04-04
Cache objects were being modified when they shouldn't

mandawilson:
Return copy of what is in cache not objects themselves.

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/model/MainType.java
+++ b/core/src/main/java/org/mskcc/oncotree/model/MainType.java
@@ -23,7 +23,12 @@ public class MainType  {
   public MainType(){}
 
   public MainType(String name) {
-      this.name = name;
+    this.name = name;
+  }
+
+  public MainType(MainType otherMainType) {
+    this.id = otherMainType.id;
+    this.name = otherMainType.name;
   }
 
   /**
--- a/core/src/main/java/org/mskcc/oncotree/model/TumorType.java
+++ b/core/src/main/java/org/mskcc/oncotree/model/TumorType.java
@@ -28,6 +28,56 @@ public class TumorType {
     private Integer level = null;
     private static final String UNKNOWN_ONCOTREE_NODE_LEVEL = "-1";
 
+    public TumorType() {}
+
+    public TumorType(TumorType otherTumorType) {
+        this.id = otherTumorType.id;
+        this.code = otherTumorType.code;
+        this.color = otherTumorType.color;
+        this.name = otherTumorType.name;
+        if (otherTumorType.mainType != null) {
+            this.mainType = new MainType(otherTumorType.mainType);
+        }
+        this.NCI = new ArrayList<String>(otherTumorType.NCI);
+        this.UMLS = new ArrayList<String>(otherTumorType.UMLS);
+        this.tissue = otherTumorType.tissue;
+        // shallow copy
+        this.children = new HashMap<String, TumorType>(otherTumorType.children);
+        this.parent = otherTumorType.parent;
+        this.deprecated = otherTumorType.deprecated;
+        this.history = new ArrayList<String>(otherTumorType.history);
+        if (otherTumorType.links != null) {
+            this.links = new Links(otherTumorType.links);
+        }
+        this.level = otherTumorType.level;
+    }
+
+    public TumorType deepCopy() {
+        TumorType newTumorType = new TumorType();
+        newTumorType.id = this.id;
+        newTumorType.code = this.code;
+        newTumorType.color = this.color;
+        newTumorType.name = this.name;
+        if (this.mainType != null) {
+            newTumorType.mainType = new MainType(this.mainType);
+        }
+        newTumorType.NCI = new ArrayList<String>(this.NCI);
+        newTumorType.UMLS = new ArrayList<String>(this.UMLS);
+        newTumorType.tissue = this.tissue;
+        newTumorType.children = new HashMap<String, TumorType>(this.children.size());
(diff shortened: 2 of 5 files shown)


zhx828:
@mandawilson why do we think these tumor type objects have been modified?

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -28,6 +28,32 @@
...
+        this.name = otherTumorType.name;
+        if (otherTumorType.mainType != null) {
+            this.mainType = new MainType(otherTumorType.mainType);
+        }
+        this.NCI = new ArrayList<String>(otherTumorType.NCI);
+        this.UMLS = new ArrayList<String>(otherTumorType.UMLS);
+        this.tissue = otherTumorType.tissue;
+        this.children = new HashMap<String, TumorType>(otherTumorType.children.size());
+        for (Map.Entry<String, TumorType> entry : otherTumorType.children.entrySet()) {
+            this.children.put(entry.getKey(), new TumorType(entry.getValue()));
```
sheridancbio: This is a recursive call .. which is a little scary in a constructor. If somebody tries to copy a node which is a descendent of itself in any way, we are headed for a stack overflow.
If we want safety from this possibility, we could do a cycle dependency check/detection ... searching if any descendant paths comes back to the current object, and if so, refusing to copy children or using a special value which represents a logical failure.

Another issue here is that every node in the cache will have multiple copies --- one for itself, plus one for every ancestor node. With a maximum tree depth of 10 the most we will have is 10 times as much memory usage as necessary. (or maybe this constructor function is only used for return values passed out of services? (rather than the cache initialization code) If that is the case, maybe this function should not be a constructor but a "cloneTumorTypeValue()" function. The constructor could do the shallow copy (re-using the children). Since we have an addChild member function, I would suggest doing the cycle detection for deep cloning. (This could be done in future PR).

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -48,14 +48,23 @@ public static void resetTumorTypesByVersion(Version version) {
...
-            return tumorTypes.get(version);
+            return getUnmodifiableTumorTypesByVersion(tumorTypes.get(version));
         }
     }
 
+    private static Map<String, TumorType> getUnmodifiableTumorTypesByVersion(Map<String, TumorType> tumorTypeMap) {
+        // code is modifying the returned tumor types, make a copy of everything
+        Map<String, TumorType> unmodifiableTumorTypeMap = new HashMap<String, TumorType>(tumorTypeMap.size());
+        for (Map.Entry<String, TumorType> entry : tumorTypeMap.entrySet()) {
+            unmodifiableTumorTypeMap.put(entry.getKey(), new TumorType(entry.getValue()));
```
sheridancbio: Ok, I see .. the clone function is for returned values from the CacheUtil. So no duplication within the cache (which uses the more primitive constructors)

mandawilson:
@zhx828 Here: https://github.com/cBioPortal/oncotree/blob/master/core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java#L190.  This modifies the tumor types in the cache, permanently breaking the tree structure.  Previously it broke the current tumor type tree, but the tree was never used again so it didn't matter.  Now that the cache works it really matters - it is why Oncotree broke this morning.

core/src/main/java/org/mskcc/oncotree/model/MainType.java
```diff
@@ -23,7 +23,12 @@
   public MainType(){}
 
   public MainType(String name) {
-      this.name = name;
+    this.name = name;
+  }
+
+  public MainType(MainType otherMainType) {
```
sheridancbio: I guess this refactor is not yet in yet .. so this modification is still needed for now

core/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -28,6 +28,56 @@
...
+        this.deprecated = otherTumorType.deprecated;
+        this.history = new ArrayList<String>(otherTumorType.history);
+        if (otherTumorType.links != null) {
+            this.links = new Links(otherTumorType.links);
+        }
+        this.level = otherTumorType.level;
+    }
+
+    public TumorType deepCopy() {
+        TumorType newTumorType = new TumorType();
```
sheridancbio: maybe a helper function private void shallowCopyAllButChildren(TumorType otherTumorType) would be helpful and allow reuse?

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -48,14 +48,23 @@ public static void resetTumorTypesByVersion(Version version) {
...
+            return getUnmodifiableTumorTypesByVersion(tumorTypes.get(version));
         } else {
             logger.debug("getTumorTypesByVersion() -- did NOT find '" + version.getVersion() + "' in cache, getting now");
             tumorTypes.put(version, TumorTypesUtil.getTumorTypesByVersionFromRaw(version));
-            return tumorTypes.get(version);
+            return getUnmodifiableTumorTypesByVersion(tumorTypes.get(version));
         }
     }
 
+    private static Map<String, TumorType> getUnmodifiableTumorTypesByVersion(Map<String, TumorType> tumorTypeMap) {
```
sheridancbio: In some ways .. since we are passing back a full, detached clone of the data it is less important that the returned map be unmodifiable. We are safe now.

PR #108, opened 2018-04-12
- changed tumor types endpoint to throw TumorTypesNotFound exception …

averyniceday:
- changed tumor types endpoint to throw TumorTypesNotFound exception when empty list is returned
- annotate possible error codes returned in endpoints
- update swagger versions in pom to 2.5.0
    - old version does not display success code
- differentiate between different responses from Crosswalk (4xx, 5xx) - throws different exceptions
- moved all exceptions/error classes into error package


web/src/main/java/org/mskcc/oncotree/api/GlobalControllerExceptionHandler.java
```diff
@@ -53,6 +49,12 @@ public void handleTopBraidException() {
         // nothing to do
     }
 
+    @ResponseStatus(code = HttpStatus.SERVICE_UNAVAILABLE, reason = "Failed to connect to CVS")
+    @ExceptionHandler(CrosswalkServiceUnavailableException.class)
+    public void handleCrosswalkServiceUnavailableException() {
```
angelicaochoa: you can make this a one liner if there's nothing to do 

`public void handleCrosswalkServiceUnavailableException() {}`

angelicaochoa: and below as well

core/src/main/java/org/mskcc/oncotree/crosswalk/CrosswalkRepository.java
```diff
@@ -52,14 +55,19 @@ public MSKConcept queryCVS(String vocabularyId, String conceptId, String histolo
         try {
             ResponseEntity<MSKConcept> response = restTemplate.getForEntity(crosswalkURL, MSKConcept.class, vocabularyId, conceptId, histologyCode, siteCode);
             return response.getBody();
-        } catch (RestClientException e) {
-            logger.error("queryCVS() -- caught RestClientException: " + e);
-            throw new CrosswalkException("Exception while getting data from CVS Service with: \n " +
-                "URI: " + crosswalkURL + "\n" +
+        } catch (HttpClientErrorException e) {
```
mandawilson: HttpClientErrorException is only one of the exceptions that might be thrown.  `restTemplate.getForEntity` throws RestClientException so you are missing exceptions like ResourceAccessException (e.g. if the URL is incorrect).  Also HttpClientErrorException is defined as "Exception thrown when an HTTP 4xx is received." so I don't think `else if (e.getStatusCode().is5xxServerError())` will ever be true.  I think this section needs to be re-written.  We can talk about alternatives if you want.

PR #110, opened 2018-04-16
Removed large margin on right side of API frame

mandawilson:


yichaoS:
Maybe a scroll bar at the bottom should also be enabled when resizing? 
screen shot 2018-04-16 at 4 33 26 pm

web/src/main/resources/static/js/news.js
```diff
@@ -26,6 +26,4 @@ function displayNews(news) {
 
 $(document).ready(function() {
     loadNews(displayNews);
-    $('#api-embed-dev').css('height', $(window).height() +'px');
```
sheridancbio: I suspect this was a typo id ... maybe should be api-embed-div ?
In index.html, this is "embed-repsonsive-16by9" class which is a bootstrap element. Maybe we should think about updating bootstrap to get better behavior?
Also, these statements tried to set width and height by pixels ... did we try to set by percentage (100%)?

mandawilson: Yes it was probably meant to be api-embed-div, but in any case it matches id='api-embed-dev' in index.html, so it was working. I deleted them because I couldn't see what it was adding, besides a large margin.  Also I meant to test the latest bootstrap but then forgot to.  I will try it now.

PR #111, opened 2018-04-18
Updated JavaScript libraries

mandawilson:
Updated bootstrap, jquery, qtip, and clipboard.  Did not update d3 or showdown.  I attempted d3, but migrating was too big a task.  The "Back to Top" does not work as well as it used to, but I believe otherwise things are mostly the same.  The button style changed slightly.

I am not sure if we should actually bother to do this update or not.


sheridancbio:
I think this is great. I'll take a look.

PR #113, opened 2018-04-20
Maint 0.0.1

sheridancbio:
Modifications for maintenance of legacy API web application:
* some bugfixes included
* some cleanup of unused classes included
* the implementation of the cache refresh functionality included (but not the full API cleanup)
* warning messages about upcoming discontinuation added


sheridancbio:
Note : this needs a fix to the refreshCache API, which got disconnected during the split-up of the API cleanup changeset.

web/src/main/java/org/mskcc/oncotree/api/GlobalControllerExceptionHandler.java
```diff
@@ -44,7 +44,7 @@
 
     private static final Logger logger = LoggerFactory.getLogger(GlobalControllerExceptionHandler.class);
 
-    @ResponseStatus(code = HttpStatus.INTERNAL_SERVER_ERROR, reason = "Failed to connect to TopBraid")
+    @ResponseStatus(code = HttpStatus.SERVICE_UNAVAILABLE, reason = "Failed to connect to TopBraid")
```
mandawilson: We probably really shouldn't be changing the behavior of the API here, but OK.

web/src/main/resources/static/index.html
```diff
@@ -41,6 +41,10 @@
           <div class="row">
             <div class="col-lg-6 col-md-8 col-sm-12 col-xs-12">
               <h3>CMO Tumor Type Tree</h3>
+              <div class="alert alert-warning alert-dismissible" role="alert">
```
mandawilson: These are nice, good idea.

web/src/main/java/org/mskcc/oncotree/api/GlobalControllerExceptionHandler.java
mandawilson: I guess it was part of the commit I had done.

sheridancbio: I'll revert this to the previous status code (INTERNAL_SERVER_ERROR) now.

sheridancbio:
Made fix for avoiding the delivery of children links during API search endpoints.

PR #121, opened 2018-05-08
Never null the cache

mandawilson:


core/src/test/java/org/mskcc/oncotree/utils/TumorTypesUtilTest.java
```diff
@@ -214,4 +220,39 @@ public void testFullTumorTypesTxt() throws Exception {
...
+        CacheUtil.setDateOfLastCacheRefresh(validDateOfLastCacheRefresh);
+        if (CacheUtil.cacheIsStale()) {
+            CacheUtil.resetCache();
+        }
+        assertThat(CacheUtil.getDateOfLastCacheRefresh().toString(), equalTo(validDateOfLastCacheRefresh.toString()));
+
+        // test condition where cache is stale
+        // set cache age to a valid age (current date - MAXIMUM_CACHE_AGE_IN_DAYS - 3)
+        // cache is stale so resetCache should be called -- cache age should change
+        currentDate.add(Calendar.DATE, -(CacheUtil.MAXIMUM_CACHE_AGE_IN_DAYS + 3));
```
sheridancbio: here we are re-using the currentDate object. But wasn't current date offset already on line 231 to be two days before "now" ? So by subtracting 6 days on line 242 don't we get to a time point 8 days before "now"?  This doesn't break what we want .. the second time point is "old enough", but if additional tests are added with new timepoints the programmer may be confused about why their times are not what they expect ... if I'm understanding Calendar properly.

PR #130, opened 2018-06-20
Hack to add CLL to history for node CLLSLL

mandawilson:


core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -84,6 +84,11 @@ private void resetCache() {
                 } else {
                     topBraidURIsToOncotreeCodes.put(node.getURI(), new HashSet<String>());
                 }
+                // TODO replace this hack for adding CLL to the history with something smarter
+                if (node.getURI().endsWith("ONC000369")) {
```
sheridancbio: This would add CLL to the history of ONC000369 across all time. But in earlier versions of the oncotree (before the "fusion" of CLL and SLL) we will have an actual oncotree node for CLL existing concurrently with ONC000369 ... which had the oncotree code of "SLL" at that point in the history. See http://oncotree.mskcc.org/api/tumor_types.txt?version=oncotree_2018_05_01
for the last point in history where this was true. So the API would output both a TumorType "CLL" and also a TumorType "SLL" with history showing "CLL". I think that would be confusing, so I request that we only add CLL to the history if the URI "ONC000044" is not present in the list of oncoTreeNodes in the version being processed. This can be done without much additional overhead if we introduce a variables "OncotreeNode fusedCllSllNode" and "OncotreeNode standaloneCllNode" both initialized to null at the start of the loop, and then set them if the appropriate URI is encountered during the loop. Then outside the loop we would add the history only if fusedCllSllNode was set and standaloneCllNode was null. (maybe I'll do a PR to your fork for this if it sounds good to you)

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
```diff
@@ -174,6 +174,7 @@ private Version setupMockVersion() {
         List<OncoTreeNode> tmpOncoTreeRepositoryMockResponse = new ArrayList<>();
         for (int pos = 0; pos < rawTestValueSource.length; pos = pos + valuesPerCase) {
             OncoTreeNode nextNode = new OncoTreeNode();
+            nextNode.setURI("TESTURI");
```
sheridancbio: Was this just an oversight during the construction of our unit tests?  I guess we need this because we are now calling getURI() and .endsWith() and would get a NPE if not set -- ok, understood.

sheridancbio:
This changeset was incorporated into #131

PR #131, opened 2018-06-25
Tweaked hack for cll code

sheridancbio:
hack for getting CLL into history of CLLSLL


sheridancbio:
correction made from comment by @averyniceday (to use full URI strings to avoid the .endsWith() potential ambiguity)

I have deployed the latest code to : http://dashi-dev.cbio.mskcc.org:8080/sheridan-oncotree/#/home

PR #133, opened 2018-06-28
VersionUtil changed to go through CacheUtil instead of directly querying TopBraid

averyniceday:
- updated oncotree VersionUtil to only interact with CacheUtil (containing versions)
- version cache is keyset of tumorTypes cache in CacheUtil
- CacheUtil initializes cache - call to getCachedVersions from VersionUtil will initialize in the case of a null cache
- new VersionUtil unit test to check VersionUtil only interacts with cache
- new CacheUtil unit test to check CacheUtil getCachedVersions()/resetCache() dependency on OncoTreeVersionRepository

Cache (CacheUtil) will be initialized on startup and can only be refreshed through resetCache() endpoint (either manually or when cache is stale). A cache refresh will abort if TopBraid is down -- meaning there should always be a valid cache in place. Queries related to oncotree versions (through VersionUtil) now check the cache (CacheUtil) instead of directly querying TopBraid (OncoTreeVersionRepository).  This breaks the dependency on TopBraid (assuming initial startup is successful).  

Cache (MSKConceptCache) will also be initialized on startup and is refreshed once a week (cron-scheduled). Refresh will abort if TopBraid is down -- valid cache will always be in place. 

Application will fail to start up if TopBraid is down because the Beans are unable to be initialized (i.e MSKConceptCache bean).

### DIFF
--- a/core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
+++ b/core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
@@ -26,7 +26,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.mskcc.oncotree.topbraid.OncoTreeNode;
 import org.mskcc.oncotree.topbraid.OncoTreeRepository;
-import org.mskcc.oncotree.utils.VersionUtil;
+import org.mskcc.oncotree.topbraid.OncoTreeVersionRepository;
+import org.mskcc.oncotree.topbraid.TopBraidException;
+import org.mskcc.oncotree.utils.FailedCacheRefreshException;
 import org.mskcc.oncotree.model.Version;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.scheduling.annotation.EnableScheduling;
@@ -54,7 +56,7 @@ public class MSKConceptCache {
     private CrosswalkRepository crosswalkRepository;
 
     @Autowired
-    private VersionUtil versionUtil;
+    private OncoTreeVersionRepository oncoTreeVersionRepository;
 
     public MSKConcept get(String oncoTreeCode) {
         if (oncoTreeCodesToMSKConcepts.containsKey(oncoTreeCode)) {
@@ -62,7 +64,10 @@ public class MSKConceptCache {
             return oncoTreeCodesToMSKConcepts.get(oncoTreeCode);
         }
         logger.debug("get(" + oncoTreeCode + ") -- NOT in cache, query crosswalk");
-        return getFromCrosswalkAndSave(oncoTreeCode);
+        MSKConcept concept = getFromCrosswalk(oncoTreeCode);
+        // save even if has no information in it
+        oncoTreeCodesToMSKConcepts.put(oncoTreeCode, concept);
+        return concept;
     }
 
     @PostConstruct // call when constructed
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -120,15 +129,14 @@ public void setDateOfLastCacheRefresh(Date date) {
 
     public void resetCache() throws FailedCacheRefreshException {
         logger.info("resetCache() -- refilling tumor types cache");
-        tumorTypes = new HashMap<>();
```
averyniceday: This was wiping out our cache regardless of whether or not we could successfully query the information from TopBraid. Cache will now only be refreshed if we managed to get something back from TopBraid.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -107,21 +126,16 @@ private void resetCache() {
                 topBraidURIsToOncotreeCodes.get(FUSED_CLL_SLL_URI).add("CLL");
             }
         }
+        oncoTreeCodesToMSKConcepts = latestOncoTreeCodesToMSKConcepts;
     }
-
-    private MSKConcept getFromCrosswalkAndSave(String oncoTreeCode) {
-        // only save if we have not seen before (UMLS/NCI info will not be different)
```
averyniceday: redundant with code on line ~62 in get() function.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -107,21 +126,16 @@ private void resetCache() {
...
-        if (oncoTreeCodesToMSKConcepts.containsKey(oncoTreeCode)) {
-            return oncoTreeCodesToMSKConcepts.get(oncoTreeCode);
-        }
+    
+    private MSKConcept getFromCrosswalk(String oncoTreeCode) {
```
averyniceday: getFromCrosswalk and saving into the cache is now separate -- can be used both for the get() function which saves directly to the existing cache or the resetCache() function.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -54,31 +56,48 @@
...
+        return concept;
     }
 
     @PostConstruct // call when constructed
     @Scheduled(cron="0 0 3 * * SUN") // call every Sunday at 3am
     private void resetCache() {
         final String STANDALONE_CLL_URI = "http://data.mskcc.org/ontologies/oncotree/ONC000044";
         final String FUSED_CLL_SLL_URI = "http://data.mskcc.org/ontologies/oncotree/ONC000369";
-        logger.info("resetCache() -- clearing Crosswalk MSKConcept cache and refilling");
-        oncoTreeCodesToMSKConcepts.clear();
```
averyniceday: not nulling out the cache before successful interaction with TopBraid

core/src/test/java/org/mskcc/oncotree/utils/VersionUtilTest.java
```diff
@@ -130,6 +144,43 @@ public void testFullVersionList() throws Exception {
...
+        config.resetNotWorkingVersionRepository(oncoTreeVersionRepository);
+        try {
+            cacheUtil.resetCache();
+        } catch (FailedCacheRefreshException e) {
+            // Do nothing - exception should be thrown, cache is nulled out
+        }
+        int currentNumberOfVersions = versionUtil.getVersions().size();
+        if (currentNumberOfVersions != oncoTreeVersionRepositoryMockResponse.size()) {
+            failureReport.append("An invalid recache attempt was recognized by VersionUtil. Oncotree will fail during TopBraid downtime. Expected number of versions: " + String.valueOf(oncoTreeVersionRepositoryMockResponse.size()) + ", Returned number of versions: " + String.valueOf(currentNumberOfVersions) + "\n");
+            failureCount = failureCount +1;
```
sheridancbio: add space between + and 1 (same comment below)

PR #137, opened 2018-07-20
disable/hide oncotree-mappings tab

sheridancbio:
- comment out oncotree-mappings tab elements
- add back download sections into comment for future use


sheridancbio:
Tested on dashi-dev by eye.

PR #138, opened 2018-07-26
integration tests to ensure oncotree release doesn't break downstream…

averyniceday:
… scripts


averyniceday:
PR #136 allowing packaging of oncotree as a jar needs to be merge first: https://github.com/cBioPortal/oncotree/pull/136

integration-tests/test-oncotree-dependent-tools.sh
```diff
@@ -0,0 +1,115 @@
...
+
+    # kill test ONCOTREE process
+    ONCOTREE_PROCESS_NUMBER=`find_oncotree_process_number $ONCOTREE_PORT`
+    kill $ONCOTREE_PROCESS_NUMBER
+fi
+ 
+rm -rf $TESTING_DIRECTORY_TEMP
+
+# all four tests must pass for integration test to succeed 
+if [[ $ONCOTREE_CODE_CONVERTER_TEST_SCCESS -eq 0 || $FAKE_ONCOTREE_VERSION_TEST_SUCCESS -eq 0 || $ONCOTREE_CODE_CONVERTER_OUTPUT_TEST_SUCCESS -eq 0 ]] ; then
```
sheridancbio: shouldn't the logic be "A and B and C" rather than "A or B or C" ?
Also, the comment says four tests, but there are three conditions being checked here.

integration-tests/verify_oncotree_code_converter.py
```diff
@@ -0,0 +1,46 @@
...
+FILE_HEADERS = linecache.getline(oncotree_code_converter_output, 1).rstrip().split("\t")
+if not all([True if header in FILE_HEADERS else false for header in REQUIRED_HEADERS]):
+    print "Error, output file is missing some of the required headers. REQUIRED: " + ', '.join(REQUIRED_HEADERS) + " RETURNED: " + ', '.join(FILE_HEADERS)
+    sys.exit(1)
+
+# checks if all returned column values are NA, if so, maybe indicate issue with returned data schema
+columns_to_check = get_required_columns(oncotree_code_converter_output)
+for column_name, column_values in columns_to_check.items():
+    if all([True if column_value == "NA" else False for column_value in column_values]):
+        print "Error, all column values for in " + column_name + " are NA, please check returned oncotree schema"
```
sheridancbio: "for in" seems like a typo

integration-tests/default_oncotree_file.txt
```diff
@@ -0,0 +1,10 @@
+SAMPLE_ID	ONCOTREE_CODE
+P-1111111-T11-IH1	MDS
+P-2222222-T22-IH2	BLL
+P-3333333-T33-IH3	PEMESO
+P-4444444-T44-IH4	MPN
+P-5555555-T55-IH5	PV
+P-6666666-T66-IH6	UNKNOWN
```
sheridancbio: I guess this is expected to generate "NA", "NA" in CANCER_TYPE and CANCER_TYPE_DETAILED .. but the test only fails if *all* values are NA. So one successful conversion from an oncotree code is adequate to pass the test. (I think that makes sense since we are checking the schema compatibility)

integration-tests/test-oncotree-dependent-tools.sh
averyniceday: the SUCCESS flag is set to 1 if the individual test succeeds (0 indicates failure). 
Using "or" here means we will exit with a non-zero exit status (failed build in Jenkins) if any of the tests fail.

integration-tests/verify_oncotree_code_converter.py
```diff
@@ -0,0 +1,46 @@
...
+# returns a dictionry from column name to list of column values
+# exits if one of the columns has empty value
+def get_required_columns(oncotree_code_converter_output):
+    with open(oncotree_code_converter_output) as file:
+        header_processed = False
+        header_index_mapping = {}
+        to_return = {header : [] for header in REQUIRED_HEADERS}
+        for line in file.readlines():
+            data = line.rstrip().split("\t")
+            if not header_processed:
```
mandawilson: You can just use Python's csv.DictReader, which will read the header and then you can access the values using the expected keys like you are below: https://docs.python.org/2/library/csv.html

integration-tests/verify_oncotree_code_converter.py
```diff
@@ -0,0 +1,46 @@
...
+                header_index_mapping = {data[index] : index for index in range(len(data))}
+                header_processed = True
+                continue
+            for header in REQUIRED_HEADERS:
+                try:
+                    to_return[header].append(data[header_index_mapping[header]])
+                except KeyError:
```
mandawilson: I don't think this will work, or rather it works just because your input file has no extra columns, you are not testing for "" values, which is what you would have if your header row looked like this "BLAH\tONCOTREE_CODE\tCANCER_TYPE\tCANCER_TYPE_DETAILED\tFOO" and your data row looked like this "junk\t\t\tmore junk".  The rstrip won't strip any tabs, and then you will split and get "" values for each of your required data fields.  I would probably just use the csv.DictReader and then test for None and "" blank values as well as the KeyErrors like you are.

integration-tests/verify_oncotree_code_converter.py
```diff
@@ -0,0 +1,46 @@
...
+                    print "Error, missing value in column: " + header
+                    sys.exit(1)
+    return to_return 
+
+# check if all required headers are in the output file
+FILE_HEADERS = linecache.getline(oncotree_code_converter_output, 1).rstrip().split("\t")
```
mandawilson: get_required_columns can take care of this test since it will have read the line anyway.

integration-tests/verify_oncotree_code_converter.py
averyniceday: this is way easier!! I should have used this in all my scripts

integration-tests/test-oncotree-dependent-tools.sh
```diff
@@ -0,0 +1,124 @@
...
+
+    # maximum time to wait for oncotree to deploy 2 minutes
+    # every 10 seconds check if job is still running
+    # attempt to hit endpoint - successful return code indicated ONCOTREE has started up
+    ONCOTREE_URL="http://dashi-dev.cbio.mskcc.org:$ONCOTREE_PORT"
+    while [ $ONCOTREE_DEPLOYMENT_SUCCESS -eq 0 ] ; do
+        CURRENT_WAIT_TIME=$(($CURRENT_WAIT_TIME + 10))
+        JOB_RUNNING=`jobs | grep "oncotree.jar" | grep "Running"`
+        if [ -z $JOB_RUNNING ] ; then
+            curl --fail -X GET --header 'Accept: */*' $ONCOTREE_URL
```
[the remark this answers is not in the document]
averyniceday: I think this might be okay as-is since we just want to know oncotree has started up (and not necessarily the response body). The integration test will check the response body of the endpoints.

This just hits the homepage repeatedly, and will return a 0 exit status (and the html page) when it finishes starting up. Before that it just returns non-zero exit status `curl: (7) couldn't connect to host`

integration-tests/test-oncotree-dependent-tools.sh
```diff
@@ -0,0 +1,132 @@
...
+# provided $ONCOTREE_PORT is defined, will find process number for process on that port and kill it
+function find_and_kill_oncotree_process {
+    ONCOTREE_PORT_NUMBER=$1
+    if [ ! -z $ONCOTREE_PORT_NUMBER ] ; then
+        ONCOTREE_PROCESS_NUMBER=`netstat -tanp | grep LISTEN | sed 's/\s\s\s*/\t/g' | grep -P ":$ONCOTREE_PORT_NUMBER\t" | cut -f6 | sed 's/\/.*//'`
+    fi
+    if [ ! -z $ONCOTREE_PROCESS_NUMBER ] ; then
+        kill -9 $ONCOTREE_PROCESS_NUMBER
+        if [ $? -gt 0 ] ; then
+            echo "failed to kill process, please check for running oncotree process"
```
sheridancbio: We could report the process number that failed to be killed too with this line.

integration-tests/test-oncotree-dependent-tools.sh
```diff
@@ -0,0 +1,132 @@
...
+
+TIME_BETWEEN_ONCOTREE_AVAILIBILITY_TESTS=3
+ONCOTREE_DEPLOYMENT_SUCCESS=0
+CURRENT_WAIT_TIME=0
+MAXIMUM_WAIT_TIME=600
+if [ $ONCOTREE_PORT -gt 0 ] ; then
+    java -jar $ONCOTREE_JAR --server.port=$ONCOTREE_PORT &
+
+    # maximum time to wait for oncotree to deploy 2 minutes
+    # every 10 seconds check if job is still running
```
sheridancbio: make this "check periodically to see if job is still running" ... it is no longer 10 seconds

PR #141, opened 2018-08-06
correct bug in integration test filename paths

sheridancbio:
the bug was that path included directory "integration-test" rather than "integration-tests"


sheridancbio:
one sec ... one more bug needs to be corrected (filename of script not found)

PR #150, opened 2019-01-25
Oncotree startup/recache succeeds w/ oncotree_latest_stable

averyniceday:
**Issue**: previous oncotree would fail to startup or recache if any oncotree version was unable to recache (whether due to invalid tree structure or TopBraid exception)

**Change**: oncotree will fail to startup or recache if `oncotree_latest_stable` cannot be constructed OR there is no valid oncotree available. In the case a different oncotree is unavailable (e.g oncotree_development), recache/startup will succeed and a slack notification will be sent about versions which could not be constructed.


core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -61,6 +62,7 @@
     private static Map<Version, Map<String, TumorType>> tumorTypes = null;
     private Date dateOfLastCacheRefresh = null;
     public static final Integer MAXIMUM_CACHE_AGE_IN_DAYS = 3;
+    public static final String REQUIRED_ONCOTREE_VERSION="oncotree_latest_stable";
```
angelicaochoa: I thought this was going to be a property?

averyniceday:
I'll update the properties file in pipelines-configuration when this is approved and merged!

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -141,11 +146,21 @@ public void resetCache() throws FailedCacheRefreshException {
...
-                throw new FailedCacheRefreshException("Failed to refresh cache");
+                failedVersions.add(version.getVersion());
+                if (version.equals(requiredOncotreeVersion)) {
+                    throw new FailedCacheRefreshException("Failed to refresh cache");
+                }
             }
         }
+        if (latestTumorTypes.keySet().size() == 0) {
+           logger.error("resetCache() -- failed to pull a single valid oncotree version");
+            throw new FailedCacheRefreshException("Failed to refresh cache");
```
angelicaochoa: there's something wrong with the spacing /tabbing here

angelicaochoa: line 158, 159 are not aligned

PR #155, opened 2019-02-25
Adding revocations and precursors properties to oncotree nodes.

mandawilson:


core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeVersionRepository.java
```diff
@@ -41,7 +41,7 @@
     private String query = "PREFIX oncotree-version:<http://data.mskcc.org/ontologies/oncotree_version/> " +
         "SELECT ?api_identifier ?graph_uri ?description ?release_date ?visible " +
         "WHERE { " +
-        "   GRAPH <urn:x-evn-master:oncotree_version> { " +
+        "   GRAPH <urn:x-evn-master:oncotree_version_testing> { " +
```
mandawilson: TODO this must be changed back.

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -306,6 +344,7 @@ private TumorType initTumorType(OncoTreeNode oncoTreeNode, Version version) thro
         tumorType.setName(oncoTreeNode.getName());
         tumorType.setColor(oncoTreeNode.getColor());
         tumorType.setParent(oncoTreeNode.getParentCode());
+        // do not copy history, rescinds, or precursors
```
sheridancbio: we should add a reason with this comment

core/src/test/java/org/mskcc/oncotree/utils/CacheUtilTest.java
```diff
@@ -61,6 +60,12 @@
     
     @Resource(name="oncoTreeAdditionalVersionRepositoryMockResponse")
     private List<Version> oncoTreeAdditionalVersionRepositoryMockResponse;
+
+    @Resource(name="legacyVersion")
+    private Version legacyVersion;
+
+    @Resource(name="latestVersion")
```
sheridancbio: we will probably need additional versions (at least one more) to test the propagation/accumulation of history over time.

core/src/test/java/org/mskcc/oncotree/utils/CacheUtilTest.java
```diff
@@ -128,4 +133,105 @@ public void testRecacheWithBrokenOncoTreeVersionRepository() {
         config.resetNotWorkingVersionRepository(oncoTreeVersionRepository);
         cacheUtil.resetCache();
     }   
+
+    private void assertNoHistoryInChildren(Map<String, TumorType> tumorTypes) {
```
sheridancbio: these three functions are very similar. we could make a more general assert which takes the field to be tested as a parameter. (same comment for the assertXInChildren functions)

core/src/test/java/org/mskcc/oncotree/utils/CacheUtilTest.java
```diff
@@ -128,4 +133,105 @@ public void testRecacheWithBrokenOncoTreeVersionRepository() {
...
+
+    @Test
+    public void testGetTumorTypesByVersion() {
+        Map<String, TumorType> returnedTumorTypes = cacheUtil.getTumorTypesByVersion(legacyVersion);
+        // there is only one node in returnedTumorTypes, and it is the TISSUE node, everything else is a child of that node
+        String rootTumorTypeCode = returnedTumorTypes.keySet().iterator().next();
+        TumorType rootTumorType = returnedTumorTypes.get(rootTumorTypeCode);
+        assertEquals("TISSUE", rootTumorTypeCode);
+        assertEquals("TISSUE", rootTumorType.getCode());
+        assertEquals(32, rootTumorType.getChildren().size());
```
sheridancbio: 32 could be a parameter defined in the config class (or wherever the mock data is) .. or be computed there.
alternatively, we could have a tree traversal to see if the tree returned matches the tree defined in our mock

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
```diff
@@ -202,20 +214,51 @@ private Version setupMockVersion() {
...
             nextNode.setMainType(rawTestValueSource[pos + 2]);
             nextNode.setColor(rawTestValueSource[pos + 3]);
             nextNode.setParentCode(rawTestValueSource[pos + 4]);
+            nextNode.setURI(rawTestValueSource[pos + 5]);
             tmpOncoTreeRepositoryMockResponse.add(nextNode);
         }
         return tmpOncoTreeRepositoryMockResponse;
     }
 
+    private List<OncoTreeNode> setupOncotreeRepositoryMockResponseHistoryChange() {
```
sheridancbio: we may want to create a general function for applying sets of modifications at particular timepoints. The destination version could be an argument to this function as well as the prior set of oncotree nodes, and only the changes intended to happen at that point in time would be applied to construct the new version.

angelicaochoa:
@mandawilson tested with importer and import types of cancer. No issues

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeVersionRepository.java
mandawilson: Done.

PR #159, opened 2019-03-09
develop new feature to open tree to a search term from url

sheridancbio:
- main.js now parses hash component of url for "search_term" parameter
- main.js now calls the search() function from tree.js on any search term passed in
- main.js hash term processing refactored and standardized. 3 recognized parameters {tab, version, search_term}
- main.js checks that tab parameter value is recognized
- fix style in main.js
- allow faster testing deployment with application.properties setting crosswalk.disable_cvs_querying

### DIFF
--- a/web/src/main/resources/application.properties-EXAMPLE
+++ b/web/src/main/resources/application.properties-EXAMPLE
@@ -9,6 +9,8 @@ topbraid.password=
 
 # for crosswalk e.g. http://data.mskcc.org/ontologies/api/concept/crosswalk?vocabularyId={vocabularyId}&conceptId={conceptId}&histologyCode={histologyCode}&siteCode={siteCode}
 crosswalk.url=
+# crosswalk.disable_cvs_querying is intended only for rapid deployment of test servers during development. If set to true, the CrosswalkRepository will never search CVS for external references (NCI, UMLS, ...)
+crosswalk.disable_cvs_querying=false
 
 # for slack
 slack.url=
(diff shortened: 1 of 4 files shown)


web/src/main/resources/static/js/main.js
```diff
@@ -18,9 +30,9 @@ $(document).ready(function(){
       $('#tree-tab').click(function (e) {
```
mandawilson: Note that this does not work: http://ONCOTREE_URL/?search_term=SEZS, but that problem was not introduced here.  You have to use: http://ONCOTREE_URL/#/home?search_term=SEZS, which is not ideal.  We should fix that another time.

web/src/main/resources/application.properties-EXAMPLE
```diff
@@ -9,6 +9,7 @@ topbraid.password=
 
 # for crosswalk e.g. http://data.mskcc.org/ontologies/api/concept/crosswalk?vocabularyId={vocabularyId}&conceptId={conceptId}&histologyCode={histologyCode}&siteCode={siteCode}
 crosswalk.url=
+crosswalk.disable_cvs_querying=
```
ao508: since the default in `CrosswalkRepository.java` is defined as `false`, I think that value should also be defined here in the example properties file with perhaps a comment indicating so

sheridancbio: change made and comment added

PR #164, opened 2019-03-12
Add log url

averyniceday:


averyniceday:
currently using test instance URL for generating links (has the new feature of opening to a specific node)

PR #168, opened 2019-03-13
Fix output file/logfile naming

averyniceday:
html file overwrites output file is file does not end in .*

for example, if output file is specified as `logfile`:
regex substitution doesn't find or replace anything, summary is written into `logfile` and overwrites actual output.

### DIFF
--- a/scripts/cross_version_oncotree_translator.py
+++ b/scripts/cross_version_oncotree_translator.py
@@ -447,8 +447,8 @@ def write_summary_file(output_file, source_version, target_version):
     ambiguous_codes = sorted([ambiguous_code for ambiguous_code, ambiguous_node in GLOBAL_LOG_MAP.items() if len(ambiguous_node[CHOICES_FIELD]) > 1], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
     resolved_codes = sorted([resolved_code for resolved_code, resolved_node in GLOBAL_LOG_MAP.items() if len(resolved_node[CHOICES_FIELD]) == 1], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
 
-
-    with open(re.sub("\.[^.]+$", "_summary.html", output_file), "w") as f:
+    html_summary_file = os.path.splitext(output_file)[0] + "_summary.html"
+    with open(html_summary_file, "w") as f:
         # General info
         f.write("<!DOCTYPE html>\n<html lang=\"en\">\n<head>\n<title>Mapping Summary</title>\n<meta charset=\"UTF-8\">\n<style>\nbody {font-family:Arial; line-height:1.4}\n\n</style>\n</head><body>\n")
         f.write("<h1>Mapping Summary</h1>\n")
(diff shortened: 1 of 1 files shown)


scripts/cross_version_oncotree_translator.py
```diff
@@ -448,7 +448,7 @@ def write_summary_file(output_file, source_version, target_version):
     resolved_codes = sorted([resolved_code for resolved_code, resolved_node in GLOBAL_LOG_MAP.items() if len(resolved_node[CHOICES_FIELD]) == 1], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
 
 
-    with open(re.sub("\.[^.]+$", "_summary.html", output_file), "w") as f:
```
ao508: @averyniceday this seems prone to bugs. Can you set some variable to the html output filename instead?

averyniceday: do you mean have a set summary html file name? like always summary_report.html?

averyniceday: @ao508 ^

ao508: @averyniceday create a variable like:

```
output_html_file = re.sub("\.[^.]+$", "_summary.html", output_file)
with open(output_html_file, 'w')
   ....
```

averyniceday: or wait do you mean do the filename construction separately outside of the open()

scripts/cross_version_oncotree_translator.py
```diff
@@ -448,7 +448,7 @@ def write_summary_file(output_file, source_version, target_version):
...
+    with open(re.sub("\.[^.]+$", "", output_file) + "_summary.html", "w") as f:
         # General info
```
mandawilson: Definitely better!  Thanks for catching @ao508 .

scripts/cross_version_oncotree_translator.py
ao508: @averyniceday if all that's being done is getting the file basename (w/o extension) then you can just 

```
output_html_file = os.path.basename(output_file).split('.')[0] + '_summary.html'
```

ao508: instead of relying on regex to do the substitution for you

ao508: @averyniceday 

you would need to add the output directory also... so perhaps:


```
output_html_file  = os.path.basename(output_file).split('.')[0] + '_summary.html'
output_file_path = os.path.join(os.path.dirname(output_file), output_html_file)
```

averyniceday: super side edge case but any chance someone passes in an output file name with multiple periods like `case.1.txt`? 
do you prefer using some split/string processing without the regex? I can use something similar

ao508: @averyniceday use this instead:

```
os.path.splitext(path)
```
>Split the pathname path into a pair (root, ext) such that root + ext == path, and ext is empty or begins with a period and contains at most one period. Leading periods on the basename are ignored; splitext('.cshrc') returns ('.cshrc', '').

see [here](https://docs.python.org/3/library/os.path.html) (scroll to bottom)

averyniceday: magic!!! okay adding it in

ao508: @averyniceday  use this:


```
html_summary_file = os.path.splitext(output_file) + '_summary.html'
```

PR #169, opened 2019-03-13
Display oncotree version converter documentation from github markdown

mandawilson:
TODO: Change URL so that it doesn't point to @ao508 's repository, but instead the cbioPortal repository.  File is: web/src/main/resources/static/js/load-markdown.js


web/src/main/resources/static/index.html
```diff
@@ -105,15 +102,13 @@ <h3>Mapping Tools</h3>
...
-                  <h5>Oncotree Version Converter</h5>
-                </div>
                 <div class="card-body">
-                  <p>The OncoTree Version Converter was developed to facilitate forwards- and backwards-mapping of OncoTree codes between OncoTree release versions.</p>
                   <p>
                     <!-- TODO change filename from cross_version_oncotree_translator.py to convert_oncotree_version.py -->
-                    <a class="btn btn-primary btn-sm" href="downloads/cross_version_oncotree_translator.py" role="button" download>Download</a> <a class="btn btn-primary btn-sm" href="http://github.com/cBioPortal/oncotree/blob/master/docs/OncoTree-Version-Converter.md" target="_blank">Documentation</a>
+                    <a class="btn btn-primary btn-sm" href="downloads/cross_version_oncotree_translator.py" role="button" download>Download OncoTree Version Converter Tool</a>
                   </p>
+                  <div id='oncotree-version-converter'>
```
ao508: one small suggestion - since there isn't anything actually in the `div`, we can make the html element self-ending I believe @mandawilson

PR #170, opened 2019-03-14
Oncotree mapper dev

sheridancbio:
- only let NA be a passthrough .. not blank and not N/A
- do not bomb on an unrecognized oncotree code .. put a note in the oncotree_code_comments and move on
- log unrecognized codes .. add a section in the summary
- make the table of contents dependent on whether sections are populated
- fix some outstanding syntax bugs
- compare forward v. backwards mapping just by comparing release date .. not sorting
- issue an error if the file has carriage returns in it
- write a message when outputing to main output file
- make headers bolder (H2 instead of H3)
- drop "you may want to .." --> "you can .."

### DIFF
--- a/docs/OncoTree-Mapping-Tool.md
+++ b/docs/OncoTree-Mapping-Tool.md
@@ -17,7 +17,7 @@ The OncoTree Mapping tool can be run with the following command:
 
 
 ```
-python <path/to/scripts/oncotree_to_oncotree.py> --source-file <path/to/source/file> --target-file <path/to/target/file> --source-version <source_oncotree_version> --target-version <target_oncotree_version> --oncotree-url <oncotree_instance_url>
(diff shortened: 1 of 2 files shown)


sheridancbio:
making changes here from change requests

sheridancbio:
from comment from @averyniceday : moving section reporting unrecognized oncotree codes to come first in the log/summary

sheridancbio:
adding check from @averyniceday so that "???" is not added more than once when computing candidate codes

sheridancbio:
it looks like we tried to switch the program argument from "output_file" to "target_file" but the parser defined arguments still has output_file defined. Checking docs to see what we claimed should be there.

sheridancbio:
The documentation page says we now take --source-file and --target-file as arguments ... so updating the script accordingly

sheridancbio:
there are other broken references ... so standardizing all references to be to source_file and target_file rather than input_file and output_file

sheridancbio:
Responding to issue raised by @averyniceday .. I tested the case of multiple occurrences of a "not found in source version" oncotree code. It was only reported once in the log/summary file, despite the code being used in three separate places in the input file.

PR #176, opened 2019-03-19
seperate single direct mapping w/ more granular choices from those w/o

averyniceday:
Resolved mappings are separated in summary file based on whether or not they have more granular codes introduced.

### DIFF
--- a/scripts/oncotree_to_oncotree.py
+++ b/scripts/oncotree_to_oncotree.py
@@ -29,6 +29,7 @@ VERSION_API_IDENTIFIER_FIELD = "api_identifier"
 VERSION_RELEASE_DATE_FIELD = "release_date"
 METADATA_HEADER_PREFIX = "#"
 PASSTHROUGH_ONCOTREE_CODE_LIST = ["NA"] # These codes will be passed through the converter without examination or alteration
+TOOL_VERSION_NUMBER = "1.1"
 
 # field names used for navigating tumor types
 CHILDREN_CODES_FIELD = "children"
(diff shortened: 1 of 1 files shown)


sheridancbio:
This change looks good, but since we are releasing a new version of this code to external users I think we better add a version number report to the output, so we can ask users which version they are running when they have problems.

Maybe we can add the number (1.1?) to the general usage statement .. or add a --version command line option to report it.

scripts/oncotree_to_oncotree.py
```diff
@@ -476,16 +477,17 @@ def write_summary_file(target_file, source_version, target_version):
     # For each category, codes with more granular codes introduced are shown first
     unmappable_codes = sorted([unmappable_code for unmappable_code, unmappable_node in GLOBAL_LOG_MAP.items() if unmappable_node[NEIGHBORS_FIELD]])
     ambiguous_codes = sorted([ambiguous_code for ambiguous_code, ambiguous_node in GLOBAL_LOG_MAP.items() if len(ambiguous_node[CHOICES_FIELD]) > 1], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
-    resolved_codes = sorted([resolved_code for resolved_code, resolved_node in GLOBAL_LOG_MAP.items() if len(resolved_node[CHOICES_FIELD]) == 1 and not ("???" in resolved_node[CHOICES_FIELD]) ], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
+    partially_resolved_codes = sorted([resolved_code for resolved_code, resolved_node in GLOBAL_LOG_MAP.items() if len(resolved_node[CHOICES_FIELD]) == 1 and not ("???" in resolved_node[CHOICES_FIELD]) and resolved_node[CLOSEST_COMMON_PARENT_FIELD]], key = lambda k: sort_by_resolution_method(k, GLOBAL_LOG_MAP[k]))
```
sheridancbio: maybe make this line and the previous line parallel ... both have a condition that the choice field not be "???" ... but one does `not ("???" in choices)` and the other does `("???" not in choices)` ... if these mean the same thing, why not use the same construction in both lines. Also, one line has the condition earlier and one has it later .. why not put this test first in both lines?

scripts/oncotree_to_oncotree.py
```diff
@@ -29,6 +29,7 @@
 VERSION_RELEASE_DATE_FIELD = "release_date"
 METADATA_HEADER_PREFIX = "#"
 PASSTHROUGH_ONCOTREE_CODE_LIST = ["NA"] # These codes will be passed through the converter without examination or alteration
+TOOL_VERSION_NUMBER="1.1"
```
sheridancbio: add spaces around assignment operator like previous lines

PR #179, opened 2019-03-20
Separate OncoTree models to a package

zhx828:


web/pom.xml
```diff
@@ -23,6 +23,7 @@
         <spring.version>4.3.8.RELEASE</spring.version>
         <spring.boot.version>1.5.3.RELEASE</spring.boot.version>
         <oncotree.core.version>0.0.1-SNAPSHOT</oncotree.core.version>
```
zhx828: I noticed the version of the core module is differ from the master. Is it intentional?

ao508: @zhx828 I doubt that was intentional. Feel free to synchronize them unless we hear otherwise from @sheridancbio or @mandawilson

zhx828: I updated the version.

sheridancbio: @zhx828 I think all modules should get the same version number ... Yes, let's synchronize. Good catch.

web/pom.xml
```diff
@@ -21,7 +21,8 @@
         <java.version>1.7</java.version>
         <spring.version>4.3.8.RELEASE</spring.version>
         <spring.boot.version>1.5.3.RELEASE</spring.boot.version>
-        <oncotree.core.version>0.0.1-SNAPSHOT</oncotree.core.version>
+        <oncotree.core.version>1.0.0-SNAPSHOT</oncotree.core.version>
```
sheridancbio: This looks good, but I think the pom for core also needs to be updated.  Unit tests failed with "Could not find artifact org.mskcc.oncotree:core:jar:1.0.0-SNAPSHOT"

zhx828: @sheridancbio ah good catch. Just updated.

zhx828:
I would consider doing a squash merge.

sheridancbio:
Yes ... I will squash and merge when the integration tests are complete

sheridancbio:
Thanks @zhx828 !

PR #181, opened 2019-03-21
created validate_oncotree_uri.py : for checking against topbraid

sheridancbio:
- topbraid polled across version history
- prior codes in use for each concept are accumulated until reaching version oncotree_candidate_release
- current version parsed and compared against S-V-O triplet statements in uri mapping file
- errors reported with appropriate exit status


integration-tests/test-oncotree-dependent-tools.sh
```diff
@@ -148,8 +150,16 @@ fi
 
 rm -rf $TESTING_DIRECTORY_TEMP
 
-# all four tests must pass for integration test to succeed
-if [[ $ONCOTREE_CODE_CONVERTER_TEST_SUCCESS -eq 0 || $FAKE_ONCOTREE_VERSION_TEST_SUCCESS -eq 0 || $ONCOTREE_CODE_CONVERTER_OUTPUT_TEST_SUCCESS -eq 0 || $ONCOTREE_VERSION_MAPPER_TEST_SUCCESS -eq 0 ]] ; then
+# test that the resource_uri_to_oncocode_mapping.txt is valid and matches Topbriad
```
ao508: typo `Topbriad` --> `Topbraid`

sheridancbio:
@ao508 : I'm a little confused. You are suggesting that the uri mapping file should be moved out of the docs directory. But this file is not just mock data for unit testing ... it is actually the persistent definition of the mapping between concept ids and oncotree codes, and is (was I guess) used by the automated tool which would generate .rdf files for import into topbraid. (https://github.com/sheridancbio/oncotree-transformer).  So this file is considered the "source of truth" to which topbraid must conform. It keeps the concept URIs consistent across different oncotree versions.
There are also external users who have been looking at this file directly to understand the history relationships. The URIs are not guaranteed to be maintained (they are not part of the API interface), but some have found them helpful to understand the internal organization of oncotree.

If you would like to see the "docs" folder become primarily about resources for frontend presentation (.md5, images, etc) that would be fine with me. But I think this file .. and perhaps other non-presented documentation about the oncotree project should be kept in a place for us to reference and to share with people using oncotree. For instance, we have a bunch of google docs that are not visible to the outside world, but if we could render those in a distributable format (.pdf? .rtf?) we could put those along with other engineering documents in a folder for our (and other developer) reference.

ao508:
@sheridancbio what if the uri mapping file was instead moved to `/src/main/resources`?

I do believe that the `data_clinical_sample_test*.txt` files however should definitely be moved to a separate directory specifically for test resources.

sheridancbio:
@ao508 ok, makes sense. We just had a conversation and decided:
 - the uri mapping file will move out of docs and into /resources
- the data_clinical_sample_test* files will drop the "_test" part of the name and move down into a directory /docs/resources
- the integration_tests data files will move down into directory integration-tests/data
I'll do these adjustments as part of this PR
The documentation for the oncotree_to_oncotree.py documentation needs to be updated for the change in test data file names and links as well.
cc: @averyniceday

sheridancbio:
- also, I notice that running the mapping tool now with the arguments specified in the tutorial section does not produce matching output to the documents checked in. The documents should be updated or the tutorial adjusted to bring them in line with each other.

sheridancbio:
Corrections have been made to render a couple of section headers correctly. Also trailing whitespace has been cleaned up.

docs/OncoTree-Mapping-Tool.md
```diff
@@ -30,163 +30,163 @@ The list of OncoTree versions available are viewable [here](http://oncotree.mskc
...
 
- **Example: Schema describing a case where OncoTree node TALL is mapped to a node with more granular children**  
- 
-![Example 3](https://raw.githubusercontent.com/averyniceday/oncotree/doc-expansion/docs/images/example_3.png)
+**The `ONCOTREE_SUGGESTIONS` column would be shown as follows:**
+> _TALL -> {TLL}, more granular choices introduced_
 
-> In `oncotree_2019_03_01`, `TALL` was a leaf node with no children. In release `oncotree_2019_06_01`, `TLL` was introduced as a replacement for `TALL` with additional children `ETPLL` and `NKCLL`. 
+### 5. Invalid Source OncoTree Code
+ An invalid souce OncoTree Code means the provided code cannot be found in the source version. In such a case, mapping cannot be attempted and the `ONCOTREE_SUGGESTSIONS` column displays the following:
```
n1zea144: Minor typos - "souce" - > "source", "SUGGESTSIONS"-> "SUGGESTIONS"

docs/OncoTree-Mapping-Tool.md
```diff
@@ -30,163 +30,163 @@ The list of OncoTree versions available are viewable [here](http://oncotree.mskc
...
-  
-The tool will output two files: `data_clinical_sample_test_remapped.txt` and `data_clinical_sample_remapped_summary.html`.  
+Download oncotree_to_oncotree.py from [here <span class="text-primary oi oi-data-transfer-download"></span>](http://oncotree.mskcc.org/downloads/oncotree_to_oncotree.py).
+#### Step 3
+Run the following command from the commandline:
+
+`python oncotree_to_oncotree.py -i data_clinical_sample.txt -o data_clinical_sample_remapped.txt -s oncotree_2018_03_01 -t oncotree_2019_03_01`
+
+The tool will output two files: `data_clinical_sample_remapped.txt` and `data_clinical_sample_remapped_summary.html`.
 #### Step 4
```
n1zea144: Lets drop the last column of all ellipses and just say "Examine data_clinical_sample.remapped.txt, its first five columns should look be as follows (with the remaining columns intact):

Maybe that will give you some breathing space between the displayed columns too.

docs/OncoTree-Mapping-Tool.md
```diff
@@ -30,163 +30,163 @@ The list of OncoTree versions available are viewable [here](http://oncotree.mskc
...
 
- ###### Sample 1  
+###### Sample 1
 
 |SAMPLE_ID|ONCOTREE_CODE  |ONCOTREE_CODE_OPTIONS|
 |:--------|:-------------:|:----------------------:|
 |S1       |	              |ALL -> {BLL,TLL}, more granular choices introduced|
 
-Source OncoTree code `ALL` maps directly to codes `BLL` and `TLL`. Users should place either `BLL` or `TLL` in the `ONCOTRE_CODE`column. The `ONCOTREE_CODE_OPTIONS` column also notes that more granular choices were introduced; as such, users can use the summary file for additional guidance. 
+Source OncoTree code `ALL` maps directly to codes `BLL` and `TLL`. Users should place either `BLL` or `TLL` in the `ONCOTRE_CODE`column. The `ONCOTREE_CODE_OPTIONS` column also notes that more granular choices were introduced; as such, users can use the summary file for additional guidance.
```
n1zea144: There is as space missing between "ONCOTREE_CODE" and "column".

docs/OncoTree-Mapping-Tool.md
```diff
@@ -30,163 +30,163 @@ The list of OncoTree versions available are viewable [here](http://oncotree.mskc
...
 The summary file provides a link to the closest shared parent node `LNM`; users can choose more granular nodes by referencing the provided tree:
-![LNM-tree](https://raw.githubusercontent.com/averyniceday/oncotree/doc-expansion/docs/images/LNM-tree.png)  
+![LNM-tree](images/LNM-tree.png)
 
-###### Sample 2  
+###### Sample 2
 
 |SAMPLE_ID|ONCOTREE_CODE  |ONCOTREE_CODE_OPTIONS|
 |:--------|:-------------:|:----------------------:|
 |S1       |	              |BALL -> {BLL}, more granular choices introduced|
```
n1zea144: Should this be "S2"?

docs/OncoTree-Mapping-Tool.md
```diff
@@ -30,163 +30,163 @@ The list of OncoTree versions available are viewable [here](http://oncotree.mskc
...
+No additional resolution is needed; the previous OncoTree code was already automatically mapped to PTCL and placed in the `ONCOTREE_CODE` column. `ONCOTREE_CODE_OPTIONS` is empty because no manual selections were necessary.
+
+###### Sample 9
 
 |SAMPLE_ID|ONCOTREE_CODE  |ONCOTREE_CODE_OPTIONS|
 |:--------|:-------------:|:----------------------:|
-|S4       |	              |HGNEE -> ???, Oncotee code not in source oncotree version|  
-  
-Source OncoTree code `HGNEE` was not found in the source OncoTree version and therefore could not be mapped. Users can either reassign a new source OncoTree code (and rerun the script) or remove the sample.  
+|S4       |	              |HGNEE -> ???, Oncotee code not in source oncotree version|
```
n1zea144: Typo - "Oncotee" -> "Oncotree"

docs/OncoTree-Mapping-Tool.md
sheridancbio: fixed

docs/OncoTree-Mapping-Tool.md
sheridancbio: done

docs/OncoTree-Mapping-Tool.md
sheridancbio: fixed

docs/OncoTree-Mapping-Tool.md
sheridancbio: yes, fixed (good catch)

docs/OncoTree-Mapping-Tool.md
sheridancbio: fixed here, and in code (this was incorrect in the mapping script too)

sheridancbio:
having anchor links within the tool documentation page is problematic. Other parts of the website use the "hash" part of the URL to open various tabs or show various oncotree versions, so the parsing of these hash tags and the configuration of the display is under javascript control now. using hashes for intra-page navigation is not automatic and we would need to program the javascript parser to do the proper navigation within this page. For now I have dropped all intra-page navigation.

The "A" element links are now shown in blue if unvisited, and black once already visited. hover-over is dark grey.
I've dropped the Additional resources section and put links in the appropriate places.

integration-tests/test-oncotree-dependent-tools.sh
sheridancbio: fixed

PR #189, opened 2019-07-02
EHCache support for OncoTree Caches

averyniceday:
- EHCache backing both MSKConceptCache and TumorTypesCache (CacheUtil)
- Includes both a standard cache + backup cache
- "Get" methods will automatically try to fallback on backup
- Update unit tests
- resetCache changed from cron-scheduler to set delay (prevents overlapping cache operations)
- removed call to OncoTreeRepository from TumorTypesUtil (instead called by CacheUtil and response passed and processed by TumorTypesUtil)
Co-authored-by: Manda Wilson <1458628+mandawilson@users.noreply.github.com>


core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
```diff
@@ -0,0 +1,289 @@
...
+            versions = oncoTreeVersionRepository.getOncoTreeVersions();
+        } catch (TopBraidException e) {
+            logger.error("Unable to get versions from TopBraid... attempting to read from backup.");
+            try {
+                versions = getOncoTreeVersionsFromPersistentCacheBackup();
+                if (versions == null) {
+                    logger.error("No versions were found in backup.");
+                    throw new RuntimeException();
+                }
+            } catch (Exception e2) {
```
ao508: same comment here regarding the try/catch block structures and throwing RuntimeException

core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
```diff
@@ -0,0 +1,289 @@
...
+    // Updating MSKConcepts -- catch cases where can't connect to Crosswalk
+    @CachePut(value = "mskConceptEHCache", key = "#oncoTreeCode", unless = "#result==null")
+    public MSKConcept updateMSKConceptInPersistentCache(String oncoTreeCode) {
+        logger.info("updating EHCache with updated MSKConcept from Crosswalk");
+        MSKConcept mskConcept = new MSKConcept();
+        try {
+            mskConcept =  crosswalkRepository.getByOncotreeCode(oncoTreeCode);
+        } catch (CrosswalkConceptNotFoundException e) {
+            return new MSKConcept();
+        } catch (CrosswalkException e) {
```
ao508: same comment here

core/src/main/java/org/mskcc/oncotree/utils/SlackUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+@Component
+public class SlackUtil {
+    private static final Logger logger = LoggerFactory.getLogger(SlackUtil.class);
+    
+    @Value("${slack.url}")
+    private String slackURL;
+    
+    public void sendSlackNotification(String message) {
+        String payload = "payload={\"channel\": \"#msk-pipeline-logs\", \"username\": \"cbioportal_importer\", \"text\": \"" + message + "\", \"icon_emoji\": \":rotating_light:\"}";
```
ao508: @averyniceday can you add the channel as a property also? That way it's not hardcoded and if we're testing / running integration tests then we can easily override the channel to the one we're supposed to use when testing or deploying test instances

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
```diff
@@ -131,6 +136,12 @@ public CrosswalkRepository crosswalkRepository() {
         return repository;
     }
 
+    @Bean
+    public SlackUtil slackUtil() {
+        SlackUtil slackUtil = Mockito.mock(SlackUtil.class);
+        return slackUtil;
```
ao508: you return `Mockito.mock(SlackUtil.class)` if there isn't any customization going on

core/src/test/resources/application.properties-EXAMPLE
```diff
@@ -0,0 +1,9 @@
+# echache configuration properties for testing
```
ao508: add slack channel name as a property please

pom.xml
```diff
@@ -28,11 +28,52 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.version>1.7</java.version>
     <tomcat.version>7.0.52</tomcat.version>
-    <spring.version>4.3.8.RELEASE</spring.version>
+    <spring.version>5.1.7.RELEASE</spring.version>
```
ao508: did we quadruple check that this spring version matches the cdd version ?

web/src/main/resources/application.properties-EXAMPLE
```diff
@@ -17,3 +17,14 @@ slack.url=
 
 # required OncoTree version for startup/recache
 required.oncotree.version=oncotree_latest_stable
+
```
ao508: this is duplicate file. Please remove the one from the core module and keep this one only.

core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
```diff
@@ -0,0 +1,289 @@
...
+
+    // retrieve cached TopBraid responses from default EHCache location
+    @Cacheable(value = "oncoTreeNodesEHCache", key = "#version.version", unless = "#result==null")
+    public ArrayList<OncoTreeNode> getOncoTreeNodesFromPersistentCache(Version version) {
+        ArrayList<OncoTreeNode> oncoTreeNodes = new ArrayList<OncoTreeNode>();
+        try {
+            oncoTreeNodes = oncoTreeRepository.getOncoTree(version);
+        } catch (TopBraidException e) {
+            logger.error("Unable to get oncotree nodes from TopBraid... attempting to read from backup.");
+            try {
```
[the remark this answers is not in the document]
averyniceday: @ao508 just tried what you suggested (renaming second exception as e) -- got this error on build: 
`variable e is already defined in method getOncoTreeVersionsFromPersistentCache()...`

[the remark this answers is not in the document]
averyniceday: but I will update the RuntimeException to include the exception object

core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
averyniceday: same as above

core/src/test/resources/application.properties-EXAMPLE
averyniceday: I think these are exclusive for unit tests so the slack channel isn't necessary

pom.xml
averyniceday: just quintuple checked!

web/src/main/resources/application.properties-EXAMPLE
averyniceday: I think the duplicate you are referencing in core is for unit tests (not all properties are needed) and one is for the actual web application (all properties required).

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -95,34 +183,34 @@ public void setDateOfLastCacheRefresh(Date date) {
         dateOfLastCacheRefresh = date;
     }
 
+    public  List<Version> getCachedVersions() {
```
sheridancbio: two spaces after public

pom.xml
sheridancbio: lol

web/pom.xml
```diff
@@ -45,7 +45,34 @@
...
+      </resource>
+    </resources>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-resources-plugin</artifactId>
+        <version>2.6</version>
+        <configuration>
+          <delimiters>
+            <delimiter>${*}</delimiter>
```
sheridancbio: I'm not familiar with this delimiter function. I think @averyniceday found a problem related to this pattern

core/src/main/java/org/mskcc/oncotree/utils/CacheUtil.java
```diff
@@ -58,35 +54,127 @@
...
+                } catch (InvalidOncoTreeDataException exception) {
+                    logger.error("Unable to get tumor types from oncotree nodes");
+                    failedVersions.add(version.getVersion());
+                    continue;
+                }
+            }
+            latestTumorTypesCache.put(version, latestTumorTypes);
+        }
+        // Fail the cache refresh if required oncotree version cannot be constructed or if no versions can be constructed
+        if (failedVersions.contains(requiredOncotreeVersion)) {
```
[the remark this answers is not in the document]
averyniceday: just checked -- looks like this was the existing behavior. We could change this in the future

core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
```diff
@@ -0,0 +1,289 @@
...
+            return new MSKConcept();
+        // can't connect to Crosswalk - attempt to fall on backup
+        } catch (CrosswalkException e) {
+            logger.error("Unable to get MSKConcept from Crosswalk... attempting to read from backup.");
+            try {
+                mskConcept = getMSKConceptFromPersistentCacheBackup(oncoTreeCode);
+                if (mskConcept == null) {
+                    return new MSKConcept();
+                }
+            } catch (Exception e2) {
```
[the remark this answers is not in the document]
ao508: an exception isn't being thrown here, my mistake. If an exception is caught at this point then a default instance of `MSKConcept` is returned.

core/src/main/java/org/mskcc/oncotree/OncoTreePersistentCache.java
ao508: @averyniceday can you fix this thrown exception as well?

core/src/main/java/org/mskcc/oncotree/utils/SlackUtil.java
```diff
@@ -0,0 +1,56 @@
...
+
+@Component
+public class SlackUtil {
+    private static final Logger logger = LoggerFactory.getLogger(SlackUtil.class);
+
+    @Value("${slack.url}")
+    private String slackURL;
+
+    @Value("${slack.channel}")
+    private String slackChannel;
```
ao508: @averyniceday the slack channel needs a "#" in front of it in the payload below. It might be a good idea to keep it hardcoded below and just have the plain channel name in the properties file. (i.e., `slack.channel=msk-pipeline-logs` instead of `slack.channel=#msk-pipeline-logs`)

ao508:
@averyniceday since the shared resources isn't something that would prohibit the oncotree war from working properly, we can add this as TODO. For now please address my second requested change (hardcoding `#` in payload for slack channel name so that the properties file contains just the plain channel name with no `#` at the beginning).

PR #216, opened 2020-09-08
Adjustments to allow use with TopBraid 6.3

sheridancbio:
- updated topbraid queries for oncotree_version
- update web request to TopBraid to include login parameter
- workaround broken MSK crosswalk API link : load current links from resource file when crosswalk.disable_cvs_querying is set
- resource file staticCrosswalkOncotreeMappings.txt added with current definitions (from production cache)
- specify key and value types for Ehcache Cache references
- switch away from deprecated ApiInfo constructor and CORS filter specification
- add additional details to logging
- drop crosswalk server interaction logic and configuration options


core/src/main/java/org/mskcc/oncotree/crosswalk/CrosswalkRepository.java
```diff
@@ -57,7 +64,7 @@ public MSKConcept getByOncotreeCode(String oncotreeCode)
     public MSKConcept queryCVS(String vocabularyId, String conceptId, String histologyCode, String siteCode)
             throws CrosswalkException {
         if (DISABLE_CVS_QUERYING) {
-            throw new CrosswalkServiceUnavailableException("CVS server queries have been disabled (crosswalk.disable_cvs_querying set to true)", new HttpServerErrorException(HttpStatus.SERVICE_UNAVAILABLE));
```
averyniceday, line 60: This is a confusing to me -- we are still attempting a query despite querying being disabled. We are then relying on a separate boolean check to throw the exception. In this case is there a reason to retain the (`DISABLE_CVS_QUERYING`) check?

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeVersionRepository.java
```diff
@@ -39,15 +39,15 @@
...
-        "   GRAPH <urn:x-evn-master:oncotree_version> { " +
-        "       ?s oncotree-version:graph_uri ?graph_uri. " +
-        "       ?s oncotree-version:api_identifier ?api_identifier. " +
-        "       ?s oncotree-version:release_date ?release_date. " +
+        "   GRAPH <urn:x-evn-master:oncotreeversionlist> { " +
+        "       ?s oncotree-version:graphUri ?graph_uri. " +
+        "       ?s oncotree-version:apiIdentifier ?api_identifier. " +
+        "       ?s oncotree-version:releaseDate ?release_date. " +
         "       OPTIONAL{?s oncotree-version:description ?description.} " +
-        "       OPTIONAL{?s oncotree-version:visible ?visible.} " +
```
averyniceday, line 50: just want to confirm the that the `oncotree-version:description` is supposed to remain `OPTIONAL` above here.

averyniceday (changes requested):
one small requested change, otherwise changes lgtm. Didn't try spinning up an instance but assuming that if unit and integration tests pass that it is okay.

averyniceday:
Integration tests didn't pass - oncotree didn't startup after 10 minutes. Did the added changes increase start time?

core/src/main/java/org/mskcc/oncotree/crosswalk/CrosswalkRepository.java
sheridancbio: I agree .. it is confusing. I guess I did not want to rip out all of the code for querying and caching crosswalk responses for the MSKConcept cache. So my thought here was to alter the meaning of crosswalk.disable_cvs_querying ... when false, an attempt is made to to query the (currently non-existent) external crosswalk server at data.mskcc.org ; when true we instead load the current production external references (NCI UMLS) from a static file.

It will be cleaner to remove the MSKConcept cache and fetch code from the codebase and the crosswalk.disable_cvs_querying configuration option. (but more work) Do you think that should be done?

averyniceday: @sheridancbio is the Crosswalk system permanently down? if so, then I think it should be removed. But if it still a viable option then I think it should be kept and this static file approach be used as a separate method (triggered through a property)

core/src/main/java/org/mskcc/oncotree/topbraid/OncoTreeVersionRepository.java
sheridancbio: Yes. When re-creating the oncotree-version ontology in a separate ontology using SHAQL constraints I thought about this and decided that maybe description could be optional, but everything else was required to be populated.

Every defined version *does* have a description. Maybe for consistency we should require all fields to be populated. Happy to change to that schema if it sounds better.

averyniceday: I don't think it's necessary :) just wanted to double check.

core/src/main/java/org/mskcc/oncotree/crosswalk/CrosswalkRepository.java
sheridancbio: @averyniceday I'm pretty sure the system is down permanently .. but I also remember that there were no updates occurring for a long time even before it went offline. I'm guessing there is no value left there, so we should trim the code and drop the property - simply using the static file for now until a new solution for external references is developed.

sheridancbio:
> Integration tests didn't pass - oncotree didn't startup after 10 minutes. Did the added changes increase start time?

I'm working on this ... I think tests are close to passing. There were a couple of problems -- one bugfix had to do with the classpath for the static resource file. Another had to do with setting the port in the integration test script. Another had to do with the short-ish wait time between probes -- when the cache is empty, each probe cause the resetCache to be restarted and it was not having enough time to ever complete successfully. Another issue is that if you cancel the tests the caches are left in an unstable state and then followup tests don't run. Hopefully I can get a successful run soon.

core/src/main/java/org/mskcc/oncotree/crosswalk/CrosswalkRepository.java
averyniceday: yes, then I think the other code should be trimmed out!

sheridancbio:
@averyniceday I have removed the configuration option for interacting with the (defunct) crosswalk server and the relevant web request logic and exception handling. I left in place (as comments) the logic for falling back on the backup cache if the service is unavailable and the primary cache is corrupted. Currently this cannot happen, because all crosswalk queries are serviced via the static resource file.

Integration tests are now running successfully (thank you @ao508). I put some code into the integration script to locate the cache and backup cache directories in a temporary space (rather than relying on the settings in the production properties file)

averyniceday (approved):

PR #217, opened 2020-09-14
fix empty crosswalk NCI references

sheridancbio:
also parameterize oncotree topbraid namespace and graph id strings


averyniceday (changes requested):
code looks good to me!  just need to update the sample application properties file and also put in default values for the injected params :)

averyniceday:
also looks like integration tests failed

sheridancbio:
application.properties.EXAMPLE updated with new properties and default values have been specified in @Value annotations.
Integration tests are now passing.

averyniceday (commented):
lgtm

PR #221, opened 2020-10-26
Update OncoTree-Mapping-Tool.md

ritikakundra:
- Updating with Ontology mapping section
- Added the ontology_to_ontology_mapping_tool


docs/OncoTree-Mapping-Tool.md
```diff
@@ -1,3 +1,7 @@
+# Contents
+* [OncoTree to OncoTree Mapping Tool](#oncotree-to-oncotree-mapping-tool)
+* [Ontology Mapping Tool](#ontology-mapping-tool)
```
sheridancbio, line 3: I suggest making the name a little more specific. To be parallel with the other tool we have, perhaps this could be called the "Ontology to Ontology Mapping tool"? Other options: "External Ontology Mapping Tool", "Cross Ontology Mapping Tool", "Inter-Ontology Mapping Tool"

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
 |P16|S16|65|MNM|
 |P17|S17|66|MYCF|
 |P18|S18|66|RBL|
+
+# Ontology Mapping Tool
+
+The Ontology Mapping tool was developed to facilitate the mapping between different cancer classification systems. We currently allow the mappings between OncoTree, ICD-O, NCIt, UMLS and HemeOnc systems. The tool and the mapping file can be found [here](https://github.com/cBioPortal/oncotree/tree/master/scripts)
```
sheridancbio, line 194: This sounds good .. but this PR should also contain the script as well (ontology_mapping_tool.py) (or .. ontology_to_ontology_mapping_tool.py) before it gets merged.

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
...
+
+### Running the tool
+
+The tool can be run with the following command:
+```
+python <path/to/scripts/ontology_mapping_tool.py> --source-file <path/to/source/file> --target-file <path/to/target/file> --target-code <target_ontology_code>
```
sheridancbio, line 200: script rename suggested above

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
...
+```
+
+**Options**
+```
+ -i | --source-file: This is the source file path. The source file must contain one of the ONCOTREE_CODE, ICDO_CODE, NCIT_CODE, UMLS_CODE or HEMEONC_CODE in the file header and it must contain codes corresponding to the Ontology System.
```
sheridancbio, line 205: What happens if the source file contains 2 or 3 of these fields? We should know what the script behavior should be and let users know here. Is there a preset order of preference? Or maybe we need a fourth command line argument "-s | --source-code" ?

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
...
+ -o | --target-file: This is the path to the target file that will be generated. It will contain ontologies mapped from source code in <source-file> to <target-code>.
+ -t | --target-code: This is the target ontology codes that the script will attempt to map the source file ontology codes to. It must be one of the ONCOTREE_CODE, ICDO_CODE, NCIT_CODE, UMLS_CODE or HEMEONC_CODE.
```
sheridancbio, line 207: here you say "codes" .. but below you say that only one code can be accepted. Maybe this should be "code"?

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
...
+```
+
+**Note:**
+- The Ontology Mapping tool runs on python 3.
+- The source file should be tab delimited with one or more columns and it should contain atleast one of ONCOTREE_CODE, ICDO_CODE, NCIT_CODE, UMLS_CODE or HEMEONC_CODE in the file header.
```
sheridancbio, line 212: I'm not sure what "one or more columns" means ... don't all non-empty source files have one or more columns? Maybe this can be dropped.  Also "atleast" --> "at least"

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
...
+- We currently are allowing only one ontology to another ontology mapping. This tool will be extended to support one to multiple ontology systems mapping.
```
sheridancbio, line 213: I would be a little less definite about extending the tool to support multiple mappings (unless we have a delivery date in mind). So I would suggest something like "In the future, we hope to extend the tool to support mapping to multiple ontology systems", or "We plan to extend the tool to support mapping to multiple ontology systems."
Also ... wouldn't you be able to run the tool 3 times in order to add three external ontologies? Maybe it is not so important to let the user do all three in one step? On the other hand, maybe it would not be too difficult to extend it so that it would do 3 at once.

sheridancbio (changes requested):
Looks good .. but a few typos and a few suggested rewordings / renamings. Also, this PR should include the mapping script itself.

docs/OncoTree-Mapping-Tool.md
ritikakundra: I like Cross ontology. Will change to that

ritikakundra: I will change it to Ontology to Ontology Mapping tool to keep it in sync with oncotree to oncotree mappint

docs/OncoTree-Mapping-Tool.md
ritikakundra: I will ask @rmadupuri to add it

docs/OncoTree-Mapping-Tool.md
ritikakundra: @sheridancbio for the time being we are only allowing one source code. The input file contains the source code. We can modify here to say just add one source code. @rmadupuri can this be updated in the read me?

docs/OncoTree-Mapping-Tool.md
ritikakundra: Agreed

docs/OncoTree-Mapping-Tool.md
rmadupuri: @sheridancbio @ritikakundra added the script and renamed it to `ontology_to_ontology_mapping_tool.py`

docs/OncoTree-Mapping-Tool.md
rmadupuri: @sheridancbio, since we are currently allowing only one to one ontology mapping, if the source file has 2 or 3 of these fields the user will be asked to enter one ontology to map by.

docs/OncoTree-Mapping-Tool.md
rmadupuri: Agreed and updated per your suggestion.

docs/OncoTree-Mapping-Tool.md
sheridancbio: Ok @rmadupuri ... so the script will exit / fail and tell the user that only one ontology column can be present in the source file?

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    if len(matched_headers) == 1:
+        source_code = matched_headers[0]
+        print("The input file has the following ontology code: %s\n" % (source_code))
+    elif len(matched_headers) == 0:
+        print("\nThe input file does not contain any ontology code columns or the headers do not match to accepted values. Please check. \n\nThe list of acceptable ontology headers are:")
+        for code in ACCEPTED_ONTOLOGIES:
+            print(code)
+        print('\n')
+        sys.exit(1)
+    else:
```
rmadupuri, line 126: @sheridancbio, heres's what the script does when it encounters more than one ontology in the source file.

docs/OncoTree-Mapping-Tool.md
rmadupuri: Yes @sheridancbio, just added another comment pointing to the block in code that handles this case.

scripts/ontology_to_ontology_mapping_tool/ontology_mappings.txt
```diff
@@ -0,0 +1,574 @@
+ONCOTREE_CODE	NCIT_CODE	UMLS_CODE	HEMEONC_CODE	ICDO_CODE
```
sheridancbio, line 1: I've scanned through this mappings file and I don't think there are any values for the ICDO_CODE column. Is that intentional?

rmadupuri: @sheridancbio yes. @ritikakundra will update the mappings soon as we get them.

docs/OncoTree-Mapping-Tool.md
```diff
@@ -184,3 +188,26 @@ After filling in the `ONCOTREE_CODE` column with an OncoTree code for each sampl
 |P16|S16|65|MNM|
 |P17|S17|66|MYCF|
 |P18|S18|66|RBL|
+
+# Ontology to Ontology Mapping Tool
+
+The Ontology Mapping tool was developed to facilitate the mapping between different cancer classification systems. We currently allow the mappings between OncoTree, ICD-O, NCIt, UMLS and HemeOnc systems. The tool and the mapping file can be found [here](https://github.com/cBioPortal/oncotree/tree/master/scripts/ontology_to_ontology_mapping_tool)
```
sheridancbio, line 194: Users do not need to download the mapping file in order to run the tool; I would make that clear here with a short statement (The mapping file does not need to be downloaded to run the tool). I now realize that in order for the user to be able to use the tool, they will need to have an internet connection so that the script can load the mapping file from github. Maybe internet connections are now an assumed part of life ... so maybe we don't need to say anything about it. Hopefully the script will output a sensible message if a user tries to run it with no network access.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
+# Copyright (c) 2019 Memorial Sloan-Kettering Cancer Center.
```
sheridancbio, line 1: 2020

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+# incidental or consequential damages, including lost profits, arising
+# out of the use of this software and its documentation, even if
+# Memorial Sloan-Kettering Cancer Center
+# has been advised of the possibility of such damage.
+
+import sys
+import os
+import io
+import argparse
+import pandas as pd
```
sheridancbio, line 21: are pandas and requests automatically built in to python3? or do they need to be installed? (I think maybe requests is not there by default). If either one is a prerequisite to running the script, we should mention it in the text on the Mapping Tool page and give an example of how to install it (pip3 install pandas .. maybe?)

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+import requests
+
+ONTOLOGY_MAPPINGS_FILE_URL = "https://raw.githubusercontent.com/cBioPortal/oncotree/master/scripts/ontology_to_ontology_mapping_tool/ontology_mappings.txt"
+VALID_ONCOTREE_CODES_URL = "https://raw.githubusercontent.com/cBioPortal/oncotree/master/resources/resource_uri_to_oncocode_mapping.txt"
+ACCEPTED_ONTOLOGIES = ['ONCOTREE_CODE','UMLS_CODE','NCIT_CODE','ICDO_CODE','HEMEONC_CODE']
+NULL_VALUES = ['','NA']
+
+def add_comments_column_and_log_data(mapped_data,target_file,source_code,target_code,source_file):
```
sheridancbio, line 29: throughout this file .. add a space character after comma in argument lists / function calls

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    completely_resolved_codes = {}
+    ambiguous_codes = {}
+    unrecognized_codes = []
+    many_to_one_codes = {}
+    mapped_codes_count = 0
+    unmapped_codes_count = 0
+
+    columns_to_groupby = mapped_data.columns[:len(mapped_data.columns)-1]
+    grouped_data = mapped_data.groupby(list(columns_to_groupby),sort=False)[target_code].unique().apply(', '.join).reset_index()
+    oncotree_codes_list = pd.read_csv(io.StringIO(requests.get(VALID_ONCOTREE_CODES_URL).content.decode('utf-8')),sep='\t',header=None,keep_default_na=False)
```
sheridancbio, line 40: why did we change "pandas" to "pd"? I think it would be more clear if we leave it as "pandas" unless everyone who uses pandas always sets it to "pd".

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    valid_oncotree_codes = oncotree_codes_list[2].str.upper().tolist()
```
sheridancbio, line 41: This is not valid. This file contains "triplets" of the form "subject verb object". This approach takes all objects as valid oncotree codes, but sometimes objects are not .. such as with the triplet
`ONC000688	hasPrecursor	ONC000360`
This says that concept ONC000688 has a precursor concept ONC000360 ... and ONC000360 is not a valid oncotree code. Maybe you could only use triplets of the form (subject hasCode object) and then use the object.
Also .. there are some rescinded oncotree codes, such as RESCINDED_URI_PREVIOUSLY_USED_FOR_PMBL. These should never occur in user files.
Also .. there are some confusing cases in the history, where oncotree codes were changed and/or rescinded. So
ONC000376	hasCode	PTCL
ONC000379	hadCode	PTCLNOS
ONC000379	hasRevocation	ONC000376
so this will take both PTCL and PTCLNOS as valid oncotree codes ... but above you give a mapping for PTCL (which is no longer a current code) and do not give a mapping for PTCLNOS (which replaced PTCL) ... there are other cases where codes were rescinded like this. I'm not sure what scenarios we want to support, but it can get complicated.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+
+    for sc,tc in zip(grouped_data[source_code],grouped_data[target_code]):
```
sheridancbio, line 43: add space after comma in tuple lists too

sheridancbio (changes requested):
partial review (up to start of script)

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+        print("\nThe input file does not contain any ontology code columns or the headers do not match to accepted values. Please check. \n\nThe list of acceptable ontology headers are:")
+        for code in ACCEPTED_ONTOLOGIES:
+            print(code)
+        print('\n')
+        sys.exit(1)
+    else:
+        print("The input file has the following ontology code columns: ")
+        for code in matched_headers:
+            print(code)
+        source_code = input("Please specify the source code column to map by: ")
```
sheridancbio, line 130: I see.. we are allowing the user to enter a choice manually to solve the problem. If this script is run as part of an automated pipeline by someone else ... then the automated pipeline will 'hang' / stall at this input statement I think. Maybe we should add a command line argument so that if a user is running in batch mode it will not try to ask the user for this input but will instead just fail at this point. Either that or we could add an "interactive" flag (-i // --interactive) which would enable the user to answer this question if it comes up. This is optional .. but if we do not do something like this, then we should mention in the README or on the Mapping Tools page that the script might ask for manual input when there are multiple source ontologies present, and so it is only intended to be run manually (and not as part of an automated pipeline).

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+        if source_code not in ACCEPTED_ONTOLOGIES:
+            print('The Source code entered is invalid.')
+            sys.exit(1)
+
+    #3. Check if source_code and target_code are not the same
+    if source_code == target_code:
```
sheridancbio, line 136: good check --- maybe another good check to add would be to see if the target ontology column already exists in the source file. The user might have a source of oncotree and a target of ICD-O ... but what would happen if the ICD-O column is already present in their source file? I suppose it would be overwritten? Maybe that is ok.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    parser.add_argument('-t', '--target-code', required = True, help = "This is the target ontology code that the script will attempt to map the source file ontology code to. It must be one of the ONCOTREE_CODE, ICDO_CODE, NCIT_CODE, UMLS_CODE or HEMEONC_CODE.",type = str)
+    args = parser.parse_args()
+
+    target_file = args.target_file
+    target_code = args.target_code.upper()
+
+    source_file = pd.read_csv(args.source_file,comment='#',sep='\t',header=0,keep_default_na=False).applymap(str) #HEMEONC_CODES are numbers.
+    source_code = extract_source_code_validate_arguments(source_file,target_code)
+
+    data = requests.get(ONTOLOGY_MAPPINGS_FILE_URL).content
```
sheridancbio, line 156: data is too general ... maybe oncotree_code_source_data ?

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    mappings_file = pd.read_csv(io.StringIO(data.decode('utf-8')),sep='\t',header=0,keep_default_na=False).applymap(str)
+
+    #For case insensitive merge.
+    source_file[source_code] = source_file[source_code].str.upper()
```
sheridancbio, line 160: there is one (old) oncotree code which had a lowercase letter ... we probably can ignore it because it is no longer current and we do not plan to allow lowercase letters again.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    mappings_file[source_code] = mappings_file[source_code].str.upper()
+
+    mapped_data = pd.merge(source_file,mappings_file[[source_code,target_code]],on=source_code, suffixes= ('_source_file',''),sort=False,how='left').fillna('')
```
sheridancbio, line 163: pandas has this merge utility ... nice

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -0,0 +1,171 @@
...
+    print("Mapped the %s to %s.." % (source_code,target_code))
+
+    add_comments_column_and_log_data(mapped_data,target_file,source_code,target_code,source_file)
+    print("\nThe ontology mappings are written to: %s" % (target_file))
+    print("The mapping summary is written to : %s" % (os.path.splitext(target_file)[0] + "_summary.html"))
```
sheridancbio, line 168: we overwrite the old summary.html file I suppose .. which is fine.

sheridancbio (changes requested):
part two of review.

Overall .. looks good 👍

docs/OncoTree-Mapping-Tool.md
rmadupuri: Updated the README to include the statement @sheridancbio

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Updated to 2020

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Good point @sheridancbio.  I have added a Prerequisites section to readme with the requirements and how to install the libraries.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Fixed

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: I have imported the pandas as "pd" only because its easier to write eachtime I need to use it :)

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Thanks for pointing it out @sheridancbio. Updated the script to use only the records with `hasCode`.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Done

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: That's right @sheridancbio. I have updated the script to read the source code as a command line argument instead of looking for it in the file. Updated the README as well.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: If the source file already has the columns, ICD-O in your example, the common columns in the source file will get the suffix `_source_file` ICDO_source_file here.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Updated

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: Yes, there was one code SpCC. I guess its no more part of the tree? And all the ontologies were uppercase in the mapping file.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
rmadupuri: I guess it's fine?

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
sheridancbio: ok, that seems reasonable. I guess if the user starts with ICDO column and then runs the script twice they may wind up with ICDO_source_file_source_file ... but we can leave it to them to clean up those columns I guess.

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
sheridancbio: 👍

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
sheridancbio: ok

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
sheridancbio: ok, looks good. There still may be some confusion and lack of support if they have not updated to the latest oncotree version .. but that is reasonable

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
sheridancbio: 👍  this works for everyone (even if it means a little extra typing)

scripts/ontology_to_ontology_mapping_tool/ontology_to_ontology_mapping_tool.py
```diff
@@ -121,21 +121,29 @@ def validate_arguments(source_code, target_code):
         print("Please specify different ontologies to map on.\n")
         sys.exit(1)
 
+    #4. CHeck if the user input file has the source code column.
+    source_file.columns = map(str.upper, source_file.columns)
+    if source_code not in source_file.columns:
+        print("\nThe input file does not contain any ontology code columns or the headers do not match to accepted values. Please check. \n\nThe list of acceptable ontology headers are:")
```
sheridancbio, line 127: sorry .. one more change request ... this massage should be "does not contain the requested source ontology code column"

sheridancbio (changes requested):

sheridancbio (approved):
nice work

PR #263, opened 2024-05-21
Switched from Topbraid backend to Graphite

mandawilson:


core/src/main/java/org/mskcc/oncotree/topbraid/TopBraidSessionConfiguration.java
```diff
@@ -1,137 +0,0 @@
...
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.context.annotation.Configuration;
-import org.springframework.http.HttpStatus;
-
-/**
- *
- * @author Manda Wilson
- **/
-@Configuration
-public class TopBraidSessionConfiguration {
```
sheridancbio, line 48: This was good / clever code which you previously wrote to handle a tricky dynamic with TopBraid. Sorry that we are discarding it here. (But I am glad we don't need to play tricks like this with graphite)

core/src/main/java/org/mskcc/oncotree/utils/OncoTreePersistentCache.java
```diff
@@ -71,16 +71,18 @@ public CacheManager getCacheManager(String ehcacheXMLFilename) throws Exception
         return cacheManager;
     }
 
-    // retrieve cached TopBraid responses from default EHCache location
+    // retrieve cached Graphite responses from default EHCache location
     @Cacheable(value = "oncoTreeNodesEHCache", key = "#version.version", unless = "#result==null")
     public ArrayList<OncoTreeNode> getOncoTreeNodesFromPersistentCache(Version version) {
+        logger.debug("getOncoTreeNodesFromPersistentCache() -- version: " + version.getVersion());
```
sheridancbio, line 77: maybe clean up if this was testing code

core/src/main/java/org/mskcc/oncotree/utils/OncoTreePersistentCache.java
```diff
@@ -71,16 +71,18 @@ public CacheManager getCacheManager(String ehcacheXMLFilename) throws Exception
...
         ArrayList<OncoTreeNode> oncoTreeNodes = new ArrayList<OncoTreeNode>();
         try {
             oncoTreeNodes = oncoTreeRepository.getOncoTree(version);
-        } catch (TopBraidException e) {
-            logger.error("Unable to get oncotree nodes from TopBraid... attempting to read from backup.");
+        } catch (GraphiteException e) {
+            logger.error("Unable to get oncotree nodes from Graphite ... attempting to read from backup.");
             try {
                 oncoTreeNodes = getOncoTreeNodesFromPersistentCacheBackup(version);
+
```
sheridancbio, line 85: this blank line does not need to be added

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -50,15 +50,14 @@ public class TumorTypesUtil {
 
     private static final Logger logger = LoggerFactory.getLogger(TumorTypesUtil.class);
     public final static String TSV_HEADER = "level_1\tlevel_2\tlevel_3\tlevel_4\tlevel_5\tlevel_6\tlevel_7\tmetamaintype\tmetacolor\tmetanci\tmetaumls\thistory";
-    private static final String TOPBRAID_BASE_URI = "http://data.mskcc.org/ontologies/oncotree/";
```
sheridancbio, line 53: 👍

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
```diff
@@ -413,6 +412,17 @@ private List<TumorType> findTumorType(TumorType allTumorTypes, TumorType current
                                 StringUtils.containsIgnoreCase(currentTumorType.getMainType(), keyword));
                 }
                 break;
+            case "clinicalcasessubset":
```
sheridancbio, line 415: This actually is not needed.
I believe that we should not expose the clinicalCasesSubset identifiers to the users. These are internal identifiers which we do not want users to use or become dependent on, so I think they should not appear in the oncotree website, or in any response from the oncotree webserver api.

So we should not support searching the oncotree based on these codes either.

core/src/test/java/org/mskcc/oncotree/utils/CacheUtilTest.java
```diff
@@ -26,8 +26,8 @@
 import org.mskcc.oncotree.error.InvalidVersionException;
 import org.mskcc.oncotree.model.TumorType;
 import org.mskcc.oncotree.model.Version;
-import org.mskcc.oncotree.topbraid.TopBraidException;
-import org.mskcc.oncotree.topbraid.OncoTreeVersionRepository;
+import org.mskcc.oncotree.graphite.GraphiteException;
```
sheridancbio, line 29: If we can change these import statements but nothing else in the rest of the code of this class, doesn't it mean that the import statements are not needed because the class names are not referenced? (at least for GraphiteException)

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
```diff
@@ -334,650 +333,650 @@ private List<String> setupExpectedMainTypeList() throws Exception {
...
-                "VULVA", "Vulva/Vagina", null, "Purple", "TISSUE", "http://data.mskcc.org/ontologies/oncotree/ONC000925",
-                "VYST", "Yolk Sac Tumor", "Germ Cell Tumor", "Purple", "VGCT", "http://data.mskcc.org/ontologies/oncotree/ONC000926",
-                "WDLS", "Well-Differentiated Liposarcoma", "Soft Tissue Sarcoma", "LightYellow", "LIPO", "http://data.mskcc.org/ontologies/oncotree/ONC000927",
-                "WDTC", "Well-Differentiated Thyroid Cancer", "Thyroid Cancer", "Teal", "THYROID", "http://data.mskcc.org/ontologies/oncotree/ONC000928",
-                "WM", "Waldenstrom Macroglobulinemia", "Non-Hodgkin Lymphoma", "LimeGreen", "BCL", "http://data.mskcc.org/ontologies/oncotree/ONC000929",
-                "WPSCC", "Warty Penile Squamous Cell Carcinoma", "Penile Cancer", "Blue", "PSCC", "http://data.mskcc.org/ontologies/oncotree/ONC000930",
-                "WT", "Wilms' Tumor", "Wilms Tumor", "Orange", "KIDNEY", "http://data.mskcc.org/ontologies/oncotree/ONC000931"};
+                "AA", "Aggressive Angiomyxoma", "Soft Tissue Sarcoma", "LightYellow", "SOFT_TISSUE", "http://data.mskcc.org/ontologies/oncotree/ONC000288", "ONC000288",
```
sheridancbio, line 336: The field value which comes before the added (final) field seems to be the URI for the concept. However, these URIs are going to be invalid under graphite (The new URIs contain UUID instead).
Maybe that entire field can be eliminated? I don't think we want to rewrite each one with the appropriate UUID.

above we are setting the URI with :
nextNode.setURI(rawTestValueSource[pos + 5]);

we should look to see if this is ever referenced in the test code. I don't see any references to:
rawTestValueSource[pos + 5] in any of the "setup" functions here. So maybe we can just drop that field.

docker/Dockerfile
```diff
@@ -7,4 +7,15 @@
...
+
+# Install autossh
+RUN apk add --no-cache autossh
+
+# Add the SSH configuration file
+RUN mkdir -p /root/.ssh
+RUN chmod 700 /root/.ssh
+RUN echo "Host *" > /root/.ssh/config && \
+    echo "    StrictHostKeyChecking no" >> /root/.ssh/config && \
+    echo "    UserKnownHostsFile=/dev/null" >> /root/.ssh/config
```
sheridancbio, line 19: I think I remember that there is a configmap we have which provide the needed credentials in a mounted filesystem
I hope autossh is resilient enough to keep this connection open. I have no experience with this tool.

model/pom.xml
```diff
@@ -6,13 +6,13 @@
   <name>OncoTree Model</name>
   <description>OncoTree Models</description>
   <artifactId>model</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
```
sheridancbio, line 9: see version comments about prior pom file

model/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -132,6 +136,19 @@ public void setMainType(String mainType) {
         this.mainType = mainType;
     }
 
+    /**
+     * Our internal id for a Onctoree concept.
+     * Do not make this a JSON property, we do not return it to the user.
+     **/
+    @JsonIgnore
```
sheridancbio, line 143: 👍 I think this means that we are keeping ccss hidden from the users

model/src/main/java/org/mskcc/oncotree/model/TumorType.java
```diff
@@ -309,6 +327,7 @@ public String toString() {
         sb.append("  color: ").append(color).append("\n");
         sb.append("  name: ").append(name).append("\n");
         sb.append("  mainType: ").append(mainType).append("\n");
+        sb.append("  clinicalCasesSubset: ").append(clinicalCasesSubset).append("\n");
```
sheridancbio, line 330: I hope this only is used for our own logging.

pom.xml
```diff
@@ -7,7 +7,7 @@
   <description>The OncoTree web application</description>
   <groupId>org.mskcc.oncotree</groupId>
   <artifactId>master</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
```
sheridancbio, line 10: see version comments about prior pom file

scripts/pom.xml
```diff
@@ -7,12 +7,12 @@
   <name>OncoTree Scripts</name>
   <description>Python Scripts</description>
   <artifactId>scripts</artifactId>
-  <version>0.1.0</version>
+  <version>2.0.0</version>
```
sheridancbio, line 10: see version comments about prior pom file

web/pom.xml
```diff
@@ -13,7 +13,7 @@
   <parent>
     <groupId>org.mskcc.oncotree</groupId>
     <artifactId>master</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>2.0.0-SNAPSHOT</version>
```
sheridancbio, line 16: see version comments about prior pom file

web/src/main/java/org/mskcc/oncotree/OncotreeApplication.java
```diff
@@ -63,7 +63,7 @@ private static ApiInfo apiInfo() {
             return apiInfoBuilder
                     .title("OncoTree API")
                     .description("OncoTree API definition from MSKCC")
-                    .version("1.0.0")
+                    .version("2.0.0")
```
sheridancbio, line 66: I believe the API is unchanged, so no change should be made here.

web/src/main/resources/application.properties.EXAMPLE
```diff
@@ -2,13 +2,13 @@ springfox.documentation.swagger.v2.path=/api-docs
...
 server.port=${port:38080}
 
-# for topbraid e.g. https://evn.mskcc.org/evn/tbl/sparql
-topbraid.url=
-topbraid.username=
-topbraid.password=
-topbraid.oncotree_namespace_prefix=
-topbraid.oncotree_version_namespace_prefix=
-topbraid.oncotree_version_list_graph_id=
+# for graphite e.g. http://localhost:7200/repositories/graphite
```
sheridancbio, line 5: 👍

sheridancbio (changes requested):
This is great work. Some changes are needed and we should talk over some parts I didn't understand.

model/src/main/java/org/mskcc/oncotree/model/Version.java
```diff
@@ -50,6 +51,9 @@ public class Version implements Serializable {
     @JsonProperty("visible")
     private boolean visible;
 
+    @JsonProperty("release_date")
+    private String releaseDate;
```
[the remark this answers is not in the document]
mandawilson: @sheridancbio This is because it was getting through in additional properties before and the way it works now it is explicitly set.  I needed a setter and a variable etc.  I did add it to the JsonPropertyOrder as you suggested.

core/src/main/java/org/mskcc/oncotree/crosswalk/MSKConceptCache.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017 - 2020 Memorial Sloan-Kettering Cancer Center.
+ * Copyright (c) 2017 - 2020, 2024 Memorial Sloan-Kettering Cancer Center.
```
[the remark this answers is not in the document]
mandawilson: I agree.

averyniceday (approved):

core/pom.xml
```diff
@@ -8,13 +8,13 @@
   <description>The core module for OncoTree</description>
   <groupId>org.mskcc.oncotree</groupId>
   <artifactId>core</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
sheridancbio: It seems like this was kept as version "2.0.0-SNAPSHOT" but I won't insist on my suggestion above.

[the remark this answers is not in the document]
sheridancbio: Oh wait .. my mistake. It was changed to 1.1.0 ... clicking on the file here (in the discussion) took me to the old/outdated version. In the full "files" view I see the change.

core/src/main/java/org/mskcc/oncotree/graphite/OncoTreeVersionRepository.java
```diff
@@ -0,0 +1,102 @@
...
+ **/
+@Repository
+public class OncoTreeVersionRepository extends GraphiteRepository<Response> {
+
+    private static final Logger logger = LoggerFactory.getLogger(OncoTreeVersionRepository.class);
+
+    @Value("${graphite.oncotree_version_namespace_prefix:http://data.mskcc.org/ontologies/oncotree-version#}")
+    private String oncotreeVersionNamespacePrefix;
+
+    @Value("${graphite.oncotree_version_list_graph_id:https://preprod3.msk.synaptica.net/concept_scheme/3bb2d189-54c5-7cf7-e44b-e44a727611e6}")
```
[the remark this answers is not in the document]
sheridancbio: This has been changed to not include the UUID

core/src/main/java/org/mskcc/oncotree/graphite/oncotree/Binding.java
```diff
@@ -0,0 +1,122 @@
...
+ * out of the use of this software and its documentation, even if
+ * Memorial Sloan-Kettering Cancer Center
+ * has been advised of the possibility of such damage.
+*/
+
+package org.mskcc.oncotree.graphite.oncotree;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class Binding {
```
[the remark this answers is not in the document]
sheridancbio: This was done (classes were moved into subpackage jsonmodeling)

core/src/main/java/org/mskcc/oncotree/utils/OncoTreePersistentCache.java
sheridancbio: dropped

core/src/main/java/org/mskcc/oncotree/utils/TumorTypesUtil.java
sheridancbio: removed

core/src/test/java/org/mskcc/oncotree/utils/CacheUtilTest.java
sheridancbio: exception import dropped

core/src/test/java/org/mskcc/oncotree/utils/OncotreeTestConfig.java
sheridancbio: field was not dropped, but was shortened to just the unique code (subset identifier)

model/pom.xml
sheridancbio: adopted

pom.xml
sheridancbio: adopted

web/pom.xml
sheridancbio: adopted

web/src/main/java/org/mskcc/oncotree/OncotreeApplication.java
sheridancbio: The change was reverted, but it seems we still kept the copyright update in place. It seems incorrect to put a copyright year (2024) update when there was not actual change to the file. But maybe we should adopt a "single unified copyright message" which is comprehensive of changes to the entire codebase. Then attach that header to every file in the codebase. If we plan to do something like that, it would save us the effort of accounting for the different change history of any particular file and how the differ from each other.

model/src/main/java/org/mskcc/oncotree/model/Version.java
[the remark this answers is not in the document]
sheridancbio: 👍

PR #285, opened 2025-01-25
fix: oncoTree rendering multiple times

NishchayRajput:
At home page the OncoTree was rendering multiple time over reaload

[Screencast from 25-01-25 02:40:50 PM IST.webm](https://github.com/user-attachments/assets/b7bd3494-eb5f-4a99-93ad-d1c95dbcf645)

It was due to the useEffect triggering multiple times. So once the tree is initialise it shouldn't be triggered again.
Video after making changes.

[Screencast from 25-01-25 02:41:23 PM IST.webm](https://github.com/user-attachments/assets/8f16b4eb-80e7-466f-b58a-313cf7d5bc97)

### DIFF
--- a/web/src/main/javascript/src/pages/Home/Home.tsx
+++ b/web/src/main/javascript/src/pages/Home/Home.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useState } from "react";
 import OncoTree, { OncoTreeNode, ToolbarAction } from "@oncokb/oncotree";
 import ToolbarItem from "../../components/Toolbar/ToolbarItem";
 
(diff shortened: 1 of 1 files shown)


NishchayRajput:
@zhx828 , @ritikakundra please review

NishchayRajput:
Please review @ritikakundra

ritikakundra:
adding @bprize15 to this

web/src/main/javascript/src/pages/Home/Home.tsx
```diff
@@ -17,21 +17,22 @@ export default function Home({
 }: IHomeProps) {
   const treeContainerRef = useRef<HTMLDivElement>(null);
   const dataRef = useRef<typeof oncoTreeData | undefined>();
-
+  const [isInitialized, setIsInitialized] = useState(false);
   useEffect(() => {
     const versionChanged = dataRef.current !== oncoTreeData;
 
-    if (treeContainerRef.current && versionChanged) {
+    if (treeContainerRef.current && versionChanged && !isInitialized) {
```
bprize15, line 24: This will disable the version change functionality so we cannot add this condition.

What you are observing is not a bug. This is due to React's [Strict Mode](https://react.dev/reference/react/StrictMode) which runs each useEffect twice in development.

However, in production, you can observe a similar bug when you navigate directly to the website without the version URL param first. This is because it fetches the default version automatically, and then again when the version is set via URL param. This is definitely worth looking into. Appreciate the effort!

bprize15 (changes requested):

PR #289, opened 2025-02-06
fix: GBASC non ascii character in NCIT code checked and corrected

### LINKED ISSUE #235: non-ascii characters in ontology mapping file
Looking at the ontology code mapping file using the github browser:
https://github.com/cBioPortal/oncotree/blob/master/scripts/ontology_to_ontology_mapping_tool/ontology_mappings.txt

there appears to be unexpected characters for the record for oncotree code GBASC in the NCIT column. This should be checked and probably corrected.

NishchayRajput:
Fixed #235


NishchayRajput:
@ritikakundra @zhx828  please review!

ritikakundra (approved):

PR #294, opened 2025-04-08
use url from domain for backend requests

bprize15:


mandawilson (approved):
Thank you!

web/src/main/javascript/src/shared/constants.ts
```diff
@@ -1,6 +1,6 @@
 import { OncoTreeNode } from "@oncokb/oncotree";
 
-export const ONCOTREE_BASE_URL = "https://oncotree.mskcc.org";
+export const ONCOTREE_BASE_URL = window.location.origin
```
bprize15, line 3: I should have updated the README. Now, the [dev server](https://vite.dev/config/server-options) will always using this URL. See `vite.config.ts`. Would you mind removing the section about updating this from the README?

PR #341, opened 2026-05-21
generate one tree per PR

bprize15:


.github/workflows/release-tree.yml
```diff
@@ -0,0 +1,94 @@
...
+        id: changed-tsv-files
+        uses: tj-actions/changed-files@9426d40962ed5378910ee2e21d5f8c6fcbf2dd96
+        with:
+          files: |
+            trees/tsv/**.txt
+
+  run-go-commands:
+    runs-on: ubuntu-latest
+    needs: check-changes
+    if: ${{ always() && needs.check-changes.outputs.has_tsv_changes == 'true' }}
```
ReshmaRamaiah10, line 29: Why do we need always() here?

.github/workflows/release-tree.yml
```diff
@@ -0,0 +1,94 @@
...
+    permissions:
+      contents: write
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
```
ReshmaRamaiah10, line 37: Why do we need `fetch-depth: 0`

.github/workflows/release-tree.yml
```diff
@@ -0,0 +1,94 @@
...
+    
+    - name: Install dependencies
+      working-directory: ./web/src/main/go
+      run: go mod download
+        
+    - name: Build tree release
+      working-directory: ./web/src/main/go
+      run: go build -o release-tree ./cmd/release-tree
+    
+    - name: Get all changed TSV files
```
ReshmaRamaiah10, line 52: Why are we doing this step again?

.github/workflows/release-tree.yml
```diff
@@ -0,0 +1,94 @@
...
+      with:
+        files: |
+          trees/tsv/**.txt
+
+    - name: Run Release Tree Script
+      working-directory: ./web/src/main/go
+      env:
+        CHANGED_FILES: ${{ steps.changed-tsv-files.outputs.all_changed_files }}
+      run: |
+        ./release-tree $CHANGED_FILES
```
ReshmaRamaiah10, line 64: Can we use `steps.changed-tsv-files.outputs.all_changed_files` directly in the run command instead of env variable?

.github/workflows/release-tree.yml
```diff
@@ -0,0 +1,94 @@
...
+          echo "No changes detected in generated files. Skipping commit."
+        fi
+      
+    - name: Build mapping validation
+      working-directory: ./web/src/main/go
+      run: go build -o validate-mappings ./cmd/validate-mappings
+
+    - name: Validate mappings
+      working-directory: ./web/src/main/go
+      run: ./validate-mappings
```
ReshmaRamaiah10, line 86: Why not use go run instead?

web/src/main/go/cmd/release-tree/main.go
```diff
@@ -16,37 +16,227 @@ import (
...
-				continue
-			} else if !errors.Is(err, os.ErrNotExist) {
-				log.Fatalf("error getting file info for '%v'", treeFilename)
-			}
+	sortedTreeFiles, err := internal.GetSortedTreeFilesWithDate()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "%v\n", err)
+		os.Exit(1)
+	}
+	mostRecentTree := sortedTreeFiles[len(sortedTreeFiles)-2] // most recent will be the one just uploaded
```
ReshmaRamaiah10: This comment seems misleading. mostRecentTree is the second-to-last tree but seems like it's the new tree that was just uploaded.

web/src/main/go/cmd/release-tree/main.go
```diff
@@ -16,37 +16,227 @@ import (
...
+		tree, err := CreateOncoTreeFromFile(file, MockPreviousCodeGetter{})
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error creating tree from %v: %v\n", file, err)
+			os.Exit(1)
+		}
+
+		treeBytes, err := json.Marshal(tree)
+		if err != nil {
+			log.Fatalf("Error marshalling tree created from %v: %v\n", file, err)
+			os.Exit(1)
```
ReshmaRamaiah10, line 42: Is `os.Exit(1)` redundant since `log.Fatalf` logs the error and exits the program with status 1.

web/src/main/go/cmd/release-tree/main.go
```diff
@@ -16,37 +16,227 @@ import (
...
+		}
+
+		err = os.WriteFile(filepath.Join(internal.TREE_FILES_PATH, jsonFilename), treeBytes, os.ModePerm)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error writing file %v: %v", jsonFilename, err)
```
ReshmaRamaiah10, line 47: Why not use  `log` to log this error to be consistent

ReshmaRamaiah10 (changes requested):

web/src/main/go/cmd/release-tree/main.go
bprize15: It's just a pattern I like better to have control of exit codes. Just my preference. Can change if you want.

web/src/main/go/cmd/release-tree/main.go
bprize15: Ah, I see. I was trying to explain that I was using `-2` because `-1` is the one just uploaded. I'll make it clearer.

.github/workflows/release-tree.yml
bprize15: This stackoverflow article explains pretty well: https://stackoverflow.com/questions/61060768/why-is-it-recommended-to-use-go-build-instead-of-go-run-when-running-a-go-ap.

Point 3 is the main use for CI/CD

.github/workflows/release-tree.yml
bprize15: Since changed files is a whitespace separated list of files, handling it using an env variable causes the action to treat it as a single argument rather than separate arguments.

.github/workflows/release-tree.yml
bprize15: I could probably combine into one big git action or pass the files along, but I think this makes the flow clearer at the cost of a pretty quick git action.

.github/workflows/release-tree.yml
bprize15: tj-actions/changed-files action needs the entire Git history to accurately calculate which files changed.

.github/workflows/release-tree.yml
bprize15: Good catch, will remove.

